Issue #5, opened 2015-09-19
[ebuild replace] please replace sanewall with firehol-2.0.3


ostree:
net-firewall/sanewall is a dead project atm, the ipv6 part of code has been moved to firehol v2 - please replace/update net-firewall/sanewall with net-firewall/firehol-2.0.3 ? 

http://firehol.org/
edit:
 https://github.com/ktsaou/firehol/issues/1

blshkv:
This is a complete opposite what you have requested 3 years ago:
https://github.com/pentoo/pentoo-historical/issues/158

ostree:
Yes, Sandwall is a fork of FireHOL with support for IPv6; In recent months developers have joined forces and integrated the changes from sanewall in fireholl v2. The aim of the exchange pentoo/pentoo-historical#158 was to support IPv6 in the livecd distribution firewall. Now the reason for the conversion is a return to the original application, which now has all the required features and is actively supported and developed. In addition, a benefit is FireQOS. "FireQOS is a traffic shaping helper. It has a very simple shell scripting language to express traffic shaping(...) " http://firehol.org/#fireqos

ostree:
...besides I tested v2 over the past few months and we can safely say that the shortcomings abandoned sanewall version (even in gentoo: https://www.mail-archive.com/gentoo-commits@lists.gentoo.org/msg108141.html) have been corrected in the new v2 edition of FireHOL.

blshkv:
ok, sounds good. I'll swap them.

blshkv:
ZC merged the pull. Thanks a lot!

PR #6, opened 2015-09-19
replace sanewall with firehol

ostree:
[ebuild replace] replace sanewall with firehol-2.0.3 https://github.com/pentoo/pentoo-overlay/issues/5


ZeroChaos-:
in the future, this could have been 1 PR

PR #44, opened 2015-10-28
ext4-crypt ebuild

ostree:
"Linux kernel 4.1 introduced native encryption for the ext4 filesystem. 
This is a userspace tool to manage encrypted ext4 directories", 
info: http://blog.quarkslab.com/a-glimpse-of-ext4-filesystem-level-encryption.html 
github: https://github.com/gdelugre/ext4-crypt 
regards

### DIFF
--- a/sys-fs/ext4-crypt/ext4-crypt-9999.ebuild
+++ b/sys-fs/ext4-crypt/ext4-crypt-9999.ebuild
@@ -1,0 +1,23 @@
+# Copyright 1999-2015 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+# $Id$
+
+EAPI=5
+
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
doesn't look like this code has been touched in 3 months; are we sure it isn't dead?

blshkv:
"At the time of writing, my personal opinion is the current kernel implementation is really shaky and unstable and should not be used."

Just use encfs for now

ostree:
I fully agree with such views, this proposal is only to evaluate new kernel features,
https://lwn.net/Articles/639427/

blshkv:
Let's not waste your work and keep it in our portage. We can add it to pentoo once it is stabilized

PR #47, opened 2015-11-12
net-analyzer/nikto: fix config file could not find error.

OwenChia:
Just found nikto can not run, and with this error:
```
- Could not find a valid nikto config file
```
The default config file is /etc/nikto.conf, but in this ebuild it installed to /etc/nikto/nikto.conf.

### DIFF
--- a/net-analyzer/nikto/nikto-2.1.5-r3.ebuild
+++ b/net-analyzer/nikto/nikto-2.1.5-r3.ebuild
@@ -40,7 +40,7 @@ src_compile() {
 }
 
 src_install() {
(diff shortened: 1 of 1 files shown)


ZeroChaos-:
thanks!

blshkv:
Please revert the commit. There is a sed line in the ebuild which fixes that:
sed -i -e 's:/etc/nikto.conf:/etc/nikto/nikto.conf:'nikto.pl

blshkv:
fixed it properly in 869671d317c220b6655eba9a4403b5801388c5b2

PR #50, opened 2015-11-16
Nessus fixes

klondi:
Fix QA_PREBUILT variable, remove unused X USE, remove non existent i386 rpm, add pax-marks to the appropriate binaries and update the quickstart instructions for nessus, remove old unused ssl symlinks.


blshkv:
Thank you, I've checked openssl dep as well. It does not depend on it indeed.

PR #60, opened 2016-01-05
dev-util/apktool-2.0.3 version bump

nyim:


blshkv:
Thanks, bumped with some cleanups: 15a975919744a5982a455b645d5b84e785bd78e8

ps. I hate git for not able to track the file and how the diff with the previous version :(

PR #88, opened 2016-04-29
bettercap: bump to v1.5.4

mudler:


ZeroChaos-:
please don't restrict mirror unless mirroring is disallowed.  yes, I know it speeds things up if they are not in the gentoo mirrors but it causes problems when I try to merge ebuilds into gentoo and have no idea why that is there.

mudler:
@ZeroChaos- sure

ZeroChaos-:
@mudler thanks!

blshkv:
Guys please do not commit it half way. You deleted the previous ebuild without adjusting keywords. That  broke the tree

PR #90, opened 2016-05-10
Prefer nmap with nse flag against lua flag in nmap_vulscan

minijackson:
I don't know if it's just a paludis bug, but my package manager wants to downgrade to nmap 6.47 (when the lua flag was still there) so I had to change the order.


blshkv:
nmap[lua] had a different meaning in previous versions and was depreciated in the latest versions.  So let's just make sure we have nmap with [nse] support. Thank you for reporting!

PR #92, opened 2016-06-01
Add ebuild for ROPgadget 5.4

tecknicaltom:
A new ebuild for ROPgadget
http://shell-storm.org/project/ROPgadget/


blshkv:
The case sensitive name of the ebuild might be against gentoo policy but let's go with it for now. Thanks!

PR #99, opened 2016-06-09
sqlmap: v1.0.6 bump

OwenChia:


blshkv:
you just just changed the date but it is still uses git repo. This is not correct. I'll fix it properly

PR #101, opened 2016-06-10
Added an ebuild for the current oclhashcat beta.

hlein:
This builds and installs fine for me, now that a couple of small
Makefile tweaks have been merged upstream.

Note, even though 3.0 is still in beta, I do not recommend trying an
ebuild for the 2.0x release sources.  The build environment in 3.0 has
improved hugely; ebuilds for building 2.0 from source are far more
painful.


blshkv:
Let's take it as a start. Thanks!

PR #102, opened 2016-06-11
net-proxy/zaproxy: v2.5.0 bump

OwenChia:
Why not include more plugins of zaproxy? e.g. portscan.


blshkv:
I think portscan is something what should be done by other tools, such as nmap.
zaproxy is a web proxy after all. You can still install it if you really want it.

OwenChia:
Yes, portscan should be done by other tools, it' fine.
My point is that how to decided a plugins should or shouldn't installed by this ebuild.

blshkv:
Initially, it was my personal choice. I was relying on what should be really useful. But it was long time and the list of plugins must be updated. So feel free to suggest a new one.

PR #103, opened 2016-06-29
Update to the 3.0.0 release (plus one cleanup).

hlein:
After 3.0.0 there was one commit that removed "this is a beta" warnings from various docs; this points at that commit rather than the release tagged commit.

### DIFF
--- a/app-crypt/oclhashcat/oclhashcat-3.0.0.ebuild
+++ b/app-crypt/oclhashcat/oclhashcat-3.0.0.ebuild
@@ -1,0 +1,25 @@
+# Copyright 1999-2016 Gentoo Foundation
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=5
+
+inherit git-2 autotools
+
+DESCRIPTION="An advanced GPU-based password recovery utility"
+HOMEPAGE="https://hashcat.net/oclhashcat/"
+EGIT_REPO_URI="https://github.com/hashcat/oclhashcat.git"
+EGIT_COMMIT="67a8d97675ce50f6692e17eeb826da604b18cd05"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~x86 ~amd64"
+DEPEND="dev-util/intel-ocl-sdk"
+
+src_prepare() {
+	#do not strip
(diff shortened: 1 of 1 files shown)


blshkv:
Looks good, thanks for keeping it up

blshkv:
actually, there is no more oclhashcat. It has been merged into hashcat. I'll make some adjustments

hlein:
Indeed!  Oh, and I see now that they've updated the main project URLs and the github repo name accordingly.

blshkv:
here it is 44222fb2a0505f9ec7da4c303978dea81b1485cf
let me know if any problem.

hlein:
Ah, there is one problem, I think it was actually present in the earlier modifications you had made, though.  The fix is simple, we need this back in:

    src_compile() {
          export PREFIX=/usr
          make
    }

Without that change, when trying to run hashcat-3.00 as a non-root user I get:

    $ hashcat -m 1500 hashfile /path/to/some.dict
    hashcat (v3.00) starting...

    ERROR: /usr/bin/hashcat.induct: Permission denied

strace'ing, it was attempting to mkdir /usr/bin/hashcat.induct

IIRC hashcat makes _compile_-time decisions about where it will put runtime files, $HOME/.hashcat/.... or  its bin dir.  It wasn't enough to set PREFIX at install time.  (And since it doesn't use ./configure, the default portage behavior isn't sufficient.)  I think this had to do with its multi-platform support, but can't remember exactly.

Meanwhile, with that `src_compile()` added, the `src_install()` that sets DESTDIR and PREFIX manually is no longer needed, the default action will work fine.

Are the above notes sufficient or would you prefer I issue a new PR that makes those changes?

Thanks!

blshkv:
That was already fixed, 334a969b9e125c650392da06252ba997e2989418
please test

hlein:
Haha, yes, that works perfectly for me, thanks :)

What I get for starting testing last night and not checking/pulling again before replying today!

blshkv:
it was an another cleanup 8 min ago edd243b4116136aec53b4c9905cff789803fca59 so yeah, sync first because we are on fire %-)

PR #107, opened 2016-07-17
ebuild: import of python-nmap from salfter's overlay

ostree:
this is one of WiFi-Pumpkin deps, could we have it in the overlay?

references:
WiFi-Pumpkin/requirements.txt
https://github.com/P0cL4bs/WiFi-Pumpkin/blob/master/requirements.txt


blshkv:
Looks good, thanks

PR #112, opened 2016-07-22
ebuild: import of net-wireless/soapysdr-9999 from stefantalpalaru's o…

ostree:
…verlay and requested net-wireless/rx_tools-9999


blshkv:
Ok, merged. Let's see if it works. Thanks!

PR #133, opened 2016-11-15
Update nessus to 6.9,1

klondi:
Also make post install messages show only on relevant install (either first install or upgrade) and add a note on the need to restart nessusd-bin after an upgrade.


blshkv:
looks good, thank you. 

There is also an old openrc script. Feel free to fix it as well.

PR #134, opened 2016-11-30
Changed libscrypt dependency to libsodium for ext4-crypt

vanscheijen:
upstream gdelugre/ext4-crypt replaced libscrypt by libsodium since Sep 25, 2016


blshkv:
Thanks!

PR #142, opened 2016-12-30
Update net-analyzer/nessus-bin to 6.9.2

klondi:


blshkv:
cool, thank you. btw, have a look at the following: https://avleonov.com/2016/11/27/fast-comparison-of-nessus-and-openvas-knowledge-bases/ It looks like openvas guys are doing good join and it might better to donate some money to them instead.

klondi:
Sadly my use of Nessus is a company requirement (read higher management) otherwise I'll happily go for OpenVAS.

blshkv:
It's better to run both ;-)

Issue #174, opened 2017-03-18
moved net-misc/openvpn


latot:
Hi!, here again!

```
emerge: there are no ebuilds to satisfy "net-misc/openvpn".
(dependency required by "net-wireless/n4p-2.2-r1::pentoo[vpn]" [installed])
(dependency required by "pentoo/pentoo-wireless-2016.3-r1::pentoo[-minipentoo]" [installed])
(dependency required by "pentoo/pentoo-2015.1::pentoo[wireless]" [installed])
(dependency required by "@selected" [set])
(dependency required by "@world" [argument])
```

now is ```net-vpn/openvpn```

Thx. Cya.

PR #175, opened 2017-03-18
fix moved packages from portage tree

latot:
Fixes https://github.com/pentoo/pentoo-overlay/issues/174

Now works emerge world.

Cya.


blshkv:
almost perfect. You need to update manifest too: "ebuild xxx-0.xx.ebuild manifest"
Thanks!

blshkv:
my bad, manifest seems fine too. Good job!

PR #187, opened 2017-05-07
unicorn: force building internal qemu w/ python 2

brianmwaters:
I am running a stock (actually hardened) Gentoo setup w/ the Pentoo overlay added later; I did not use the Pentoo installer. I also haven't enabled the pentoo profile. This might be why I'm seeing issues other users aren't.

Unicorn comes with an included version of QEMU, and it's build breaks when Python 3 is used. This patch forces it to build w/ Python 2 and adds a build dependency.

Speaking of which, I noticed that in this package the run-time dependencies include pkg-config, and the build dependencies include glib. Is that a typo? Should we swap them?

### DIFF
--- a/dev-util/unicorn/unicorn-1.0.1.ebuild
+++ b/dev-util/unicorn/unicorn-1.0.1.ebuild
@@ -17,7 +17,8 @@ IUSE_UNICORN_TARGETS="x86 m68k arm aarch64 mips sparc"
 use_unicorn_targets=$(printf ' unicorn_targets_%s' ${IUSE_UNICORN_TARGETS})
 IUSE+=" ${use_unicorn_targets}"
 
-DEPEND="dev-libs/glib:2"
+DEPEND="dev-libs/glib:2
+	dev-lang/python:2.7"
 RDEPEND="${DEPEND}
 	virtual/pkgconfig"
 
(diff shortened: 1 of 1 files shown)


blshkv:
Hi, thank you. this is a hack, but I do not see a better option for now. qemu has a proper python flag which has to be specified.
See app-emulation/qemu for example.

The same stands for build dependencies, I copied it from qemu ebuild. It must be reason in this order.

brianmwaters:
Thanks.

If you look here https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/qemu/qemu-2.8.1-r2.ebuild#n405
it looks like the QEMU ebuild basically does the same thing. There's
probably a better way than the $(which python2.7) hack that I did; I
just don't know Portage that well.

blshkv:
yeah, you have to introduce PYTHON_COMPAT, distutils-r1 and friends which *might* require some other changes. I have no time for it. Feel free to create a better pull request, I will be more than happy to merge a proper fix.

brianmwaters:
Thanks for the info, I'll look into it.

I'm liking Pentoo so far and want to turn it into something that works
well for doing CTF's out of the box, so expect more patches from me on
Capstone, Unicorn, etc. and eventually even some new tools.

blshkv:
That sounds fantastic! I use pentoo for daily pt and that area is in a good shape. Zero Chaos is an expert in wireless / radio. CTF can be yours  :)

PR #191, opened 2017-05-16
net-analyzer/nessus-bin: bump to version 6.10.5

klondi:


blshkv:
Please submit to Gentoo. They have a version ahead of us.

blshkv:
ok, so Gentoo has this version already, so I'm closing the issue

PR #197, opened 2017-05-26
net-wireless/gr-elster New Package GNU Radio block …

ephemer0l:
for decoding packets transmitted by Elster R2S smart meters


blshkv:
cheers!

PR #198, opened 2017-05-26
net-wireless/gr-(acars2,ais,dsd,elster,keyfob,lte) fix paths, USE, and add Manifests

ephemer0l:
We do it nice, 'cause we do it twice. Sorry for the needed cleanup :'-(


blshkv:
no problem at all, keep it coming ;-)

PR #202, opened 2017-05-30
Update cisco-torch-0.4b.ebuild

rain0r:
Changed capitalization of Net-SSH-Perl (see: https://github.com/gentoo/gentoo/blob/master/profiles/updates/1Q-2016#L29 )


blshkv:
Thanks!

PR #213, opened 2017-06-09
dex2jar: fix install

brianmwaters:
dex2jar reinstalled fine (when files were already in place), but had
problems on fresh install.

This is because it attempted to glob over /opt/something/*, which resolved to the string "/opt/something/*" on a fresh install (because /opt/something didn't exist), but worked on systems that already had dex2jar installed. (See the patch.) Solution was to change it to "${D}/opt/something"/*.


brianmwaters:
Let me know if you prefer that I open an issue before submitting PR's for small stuff like this.

blshkv:
I can see the problem, but your patch creates invalid symlinks: ```/var/tmp/portage/dev-util/dex2jar-2.1_pre20150601/image//opt/dex2jar/*.sh```

Could you fix it properly please?

blshkv:
ok, it is fixed now, thanks!

P.S. A pull request is fine, let's not create corporate bureaucracy  ;-)

PR #218, opened 2017-07-13
h2: remove package in favor of in-tree ebuild

fkmclane:
Since `dev-python/hyper-h2` has been in-tree for a while now, we should migrate stuff to that ebuild. This pull request puts in an update for the move and fixes dependencies for `net-proxy/mitmproxy` and `net-wireless/wifi-pumpkin`.


blshkv:
cool, done. I did not find it initially probably. Also, you forgot to remove the old ebuild. I'll do it.

PR #220, opened 2017-07-18
kaitaistruct: fix description and use preferred notation for SRC_URI

fkmclane:
The Python package `kaitaistruct` is *not* the parser generator. In mitmproxy, the parser is already generated and this package provides the runtime library the generated parser requires. Additionally, the preferred notation for `SRC_URI` for PyPI is `mirror://pypi/${PN:0:1}/${PN}/${P}.tar.gz`. I updated the description to distinguish the package and updated the source uri.

Let me know if you would like for me to change anything else or have any questions.

Thanks for all of your work!


blshkv:
We have copy pasted thw title from somewhere most probably. No idea how it went wrong. Cheers

PR #224, opened 2017-07-27
parseconfig: fix homepage, license, and lack of tests

fkmclane:
I'm not sure where you got that homepage and license, but this PR fixes those. Also, there are no distributed tests in the gem so we should indicate that.

Thanks for all of your work!


fkmclane:
I changed the URL to https.

blshkv:
cool, thanks!

PR #228, opened 2017-08-09
mitmproxy: fix tornado dependency and update -9999 dependencies

fkmclane:
The tornado dependency for mitmproxy needs an upper bound. I added that and went ahead and updated the -9999 dependencies. This required a newer pyasn1 than what is in the Gentoo tree so I added that as well.

Thanks,
Foster


blshkv:
Please do not touch dev-python/cryptography for the stable version. This is work around due to sslyze requirements,  =dev-python/cryptography-1.9

fkmclane:
I did not notice that. I'll push the change.

blshkv:
thanks!

blshkv:
fyi, you shouldn't have bumped pyasn1 with stable keywords. It screwed a stable mitimproxy requirements

PR #230, opened 2017-08-10
unicorn: fix multilib issue on install

brianmwaters:
unicorn was failing emerge's multilib-strict check (after compile; before install) on my system, which means it was trying to put files into /usr/lib instead of /usr/lib32 or /usr/lib64.


blshkv:
this likely will not work. Please add ```inherit multilib``` to the ebuild as well

brianmwaters:
Added. Thanks for the tip. Admittedly I do not understand the way portage handles multilib stuff very well.

Builds and installs fine on my machine.

blshkv:
Thanks!

PR #231, opened 2017-08-10
unicorn: two tiny changes

brianmwaters:
I "properly" fixed a hack I had done a few months ago, according to blshkv's advice: https://github.com/pentoo/pentoo-overlay/pull/187. I tested it and it builds properly even when the system python is 3.4.

Also, I added static libs; I don't see why these were not included in the first place since they're built by default w/ Unicorn.

Trying to squash PR's together now to not create overload. Although I do have one more major one coming for unicorn and unicorn-bindings, so I'll let that one wait.

### DIFF
--- a/dev-util/unicorn/unicorn-1.0.1-r2.ebuild
+++ b/dev-util/unicorn/unicorn-1.0.1-r2.ebuild
@@ -3,7 +3,8 @@
 
 EAPI=6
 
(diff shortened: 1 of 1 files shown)


blshkv:
For the static it would be good to make it optional via a static flag. That's commonly used in Gentoo and that's why it wasn't done before

PR #232, opened 2017-08-10
unicorn: add static-libs use flag

brianmwaters:
I went for static-libs, as this seems to be targeted at packages which install static libs alongside shared libs, which is what is happening here (see https://github.com/pentoo/pentoo-overlay/pull/231).


blshkv:
Great, cheers

PR #233, opened 2017-08-11
Revert "Revert "unicorn: add arm and aarch64 to base profile""

brianmwaters:
I am really sorry for all the little commits, and now literally reverting a revert commit. Turns out the issue that I initially reverted 1982f3b8ded54e3849c07dc64f4519de423b12cb was caused by ccache being misconfigured on my machine, _not_ something that should affect other users.

Looking at the logs it looks like ZeroChaos is having some trouble today too, hah ;)

Let me know if you'd rather have me pile up changes into larger patches before submitting PR's.

### DIFF
--- a/profiles/pentoo/arch/amd64/make.defaults
+++ b/profiles/pentoo/arch/amd64/make.defaults
@@ -6,4 +6,6 @@ FCFLAGS="${CFLAGS}"
 QEMU_SOFTMMU_TARGETS="arm aarch64 i386 x86_64"
 QEMU_USER_TARGETS="arm aarch64 i386 x86_64"
 
+UNICORN_TARGETS="arm aarch64 x86"
+
 GRUB_PLATFORMS="coreboot efi-32 efi-64 emu multiboot pc qemu"


blshkv:
wait a moment.  Don't we configure it in an each profile ```/profiles/pentoo/arch/XXX/package.use``` individually? That global setting does not make sense. You do not need arm on PC and so on?

brianmwaters:
Ah, I see.

My thinking was, the pentoo profiles are pretty aggressive about making sure tools have support for various file formats, etc. I figured it made sense to have Unicorn build by default with support for at least both x86 and ARM. But it did it very wrongly. I'll just leave this alone for now, thanks.

blshkv:
it is probably confusing, but this variable is NOT a supported target. It sets UNICORN_ARCHS which is basically a cross-compiling option.

blshkv:
https://github.com/unicorn-engine/unicorn/blob/master/docs/COMPILE-NIX.md
```
UNICORN_ARCHS: specify list of architectures to compiled in.
```

brianmwaters:
Hm...... I see what you mean.

I actually have it installed on my machine at the moment with UNICORN_TARGETS='arm aarch64 x86 mips', but I don't have ARM or MIPS cross compilers. So that's why I'm thinking it's a target list similar to QEMU_USER_TARGETS.

I could be totally wrong, but I don't think so. I'll investigate further and get back to you.

blshkv:
sounds good, please double check. Because I've got an impression that the tool will support all arches by default regardless of this option.

blshkv:
I have done a quick test, it looks like we should active all these by default

with x86 only:
```
./sample_arm64 
Emulate ARM64 code
Failed on uc_open() with error returned: 2 (Invalid/unsupported architecture (UC_ERR_ARCH))
````

with arm aarch64 x86 arches:
```
./sample_arm64 
Emulate ARM64 code
>>> Tracing basic block at 0x10000, block size = 0x4
>>> Tracing instruction at 0x10000, instruction size = 0x4
>>> Emulation done. Below is the CPU context
>>> X11 = 0x9abc
```

I will be waiting for you to confirm it as well

brianmwaters:
I did essentially the same exact test as well just now (except I compared compiling it w/ and w/o x86 support, and attempting to emulate x86 code), and got essentially the same exact result as you.

So I agree we should activate more/all architectures. My initial commit just added arm and aarch64 (to match pentoo's QEMU_USER_TARGETS); but if it were me, I'd add at least also MIPS. Is there any reason not to just enable all 6?

blshkv:
no reason at all, the more the better. In fact, they all activated by default (if we don't provide that variable) according to the manual. I guess we've been too flexible here ;-)

I'll fix that.

brianmwaters:
Heh, great.

By the way, unicorn-bindings will have to be changed to respect this as well, since emerging unicorn-bindings with the python USE flag actually causes the whole QEMU subsystem to be recompiled (once for unicorn and once again for unicorn-bindings). Incidentally capstone-python also has a similar problem but that's another story.

Actually, unicorn-bindings is not up-to-date and needs a bunch of work. (I had problems building the Java, Ruby and Go bindings when I made a 1.0.1 ebuild for myself.) My recommendation is to drop the ruby, java and go USE flags. After looking at the build system, I think they need changes upstream if we want an ebuild that is not really brittle. And I don't think many people use these bindings, anyway.

I'm not sure how to resolve the problem whereby the QEMU subsystem is built twice, without just integrating the unicorn and unicorn-bindings packages together (e.g, by adding a python USE flag to unicorn and making unicorn-bindings just an empty stub for unicorn[python]).

By the way, all this Unicorn work is so that I can get pwntools and angr into Pentoo.

blshkv:
feel free to file an upstream bug/pull request. They are willing to fix and fast to respond usually

PR #241, opened 2017-09-11
unicorn: add python bindings

brianmwaters:
The unicorn-bindings package has some problems: first the ebuild breaks when trying to install some of the non-Python bindings (which no packages in Pentoo depend on anyway), and second, installing the unicorn-bindings packages causes the entire Unicorn shared library to be rebuild, except that this build does not respect UNICORN_TARGETS. So as a fix, I simply added a python use flag to the unicorn ebuild, which will install the python bindings during the install. This avoids rebuilding all of Unicorn. If this is acceptable, we could potentially move forward w/ unicorn-bindings[python] being a simple wrapper for unicorn[python], if it's needed for backward-compatability.

### DIFF
--- a/dev-util/unicorn/unicorn-1.0.1-r3.ebuild
+++ b/dev-util/unicorn/unicorn-1.0.1-r3.ebuild
@@ -3,8 +3,8 @@
 
 EAPI=6
 
-PYTHON_COMPAT=( python2_7 )
-inherit multilib python-single-r1
+PYTHON_COMPAT=( python2_7 python3_4 )
+inherit multilib distutils-r1
 
 DESCRIPTION="A lightweight multi-platform, multi-architecture CPU emulator framework"
 HOMEPAGE="http://www.unicorn-engine.org"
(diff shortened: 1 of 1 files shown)


blshkv:
Should be ok, thanks. Unicorn is complex ebuild and it was not easy to make everything work from the very beginning

blshkv:
I've looked at this pull request and do not like it. Basically, you have forked python bindings. Now we have two packages which is confusing. Can you fix unicorn-bindings instead?
The reason why it is a separate ebuild is to simplify the unicorn ebuild.

brianmwaters:
I could; the problem is that, even if we made unicorn-bindings respect UNICORN_TARGETS, it would still build a completely separate libunicorn.so for installation under /usr/lib/python2.7/, which takes some extra time and feels dirty to me. It also won't automatically track changes we make to dev-util/unicorn.

blshkv:
ok, so I removed that artificial build requirement. There is absolutely no reason why the Makefile of python should compile it. Also see that bug 869. The current design is wrong and @rhelmot can not justify why it cannot be fixed other than "I forget" and "it worked for a very long time, do not touch anything or it might break."

rhelmot:
- that is absolutely not what I said
- I already said I will add an option to make the python library behave like a normal library for when you want to use it with a global unicorn install

PR #252, opened 2017-10-21
johntheripper: add -9999 ebuild

fkmclane:
This is the -9999 ebuild I've been using in my overlay. CUDA support has been removed for a while now so I took that out of the ebuild. I also took out the minimal use flag, but I can restructure this and add that back in if you feel it is necessary.

Any comments/does this look good?


blshkv:
Looks good enough for a life ebuild. Thanks!

PR #256, opened 2017-12-04
sipvicious: switch to git

brianmwaters:
ebuild was broken because they're not distributing tarballs anymore; this caused the pentoo-voip metapackage to not be able to be installed


blshkv:
well done, cheers!

PR #265, opened 2018-01-21
create_ap: add version 0.4.6

fkmclane:
Is there an interest to add this ebuild taken from the Arch package create_ap?

https://github.com/oblique/create_ap


blshkv:
I don't see why not. Thanks

PR #266, opened 2018-01-21
python-ntlm: fix description, use flags, and dependencies

fkmclane:
This package needs a proper description as well as a fix where the correct flag for examples is USE=examples (https://packages.gentoo.org/useflags/examples).


blshkv:
Great, thanks

PR #269, opened 2018-02-06
Review: Mitmproxy3

ZeroChaos-:


dev-python/h11/h11-0.7.0.ebuild
```diff
@@ -0,0 +1,21 @@
...
+SRC_URI="mirror://pypi/${PN:0:1}/${PN}/${P}.zip"
+
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+RESTRICT="test"
+
+RDEPEND=""
+DEPEND="${RDEPEND}"
```
ZeroChaos-: don't declare empty deps, especially since in this case you are overriding the default python deps which are missing from this ebuild.

ZeroChaos-:
update for python 3.6 please

blshkv:
python 3.6 is not stable yet, so I'm merging it first. Will fix the rest as we go.

PR #270, opened 2018-02-06
Nikto

ZeroChaos-:
In general, we keep comments around for a reason.  In this case, I don't see such a reason so fine with dropping them, but the bytes aren't expensive and that's why we leave a lot of them.

Ack, Anton?


blshkv:
There were an official option to use system net-libs/libwhisker but it is broken. So I added a reminder that it cannot be used. Please add it back.

PR #275, opened 2018-03-17
pentoo-updater: bashism fix

Scall:


blshkv:
sorry, what's wrong with the current code?

blshkv:
It is sequence vs grouping. All gentoo ebuilds use {}, see python_compat variable.

Rejected until you provide a good explanation

Scall:
"{3,4,5}" is a bashism, therefore it won't work as expected when /bin/sh is a symlink to a POSIX-compliant shell that is not bash. You can use it in ebuilds because they are bash scripts, but you shouldn't use it in scripts with the #!/bin/sh shebang (instead of #!/bin/bash). It won't work with dash, etc.

ZeroChaos-:
I'm running out the door but I'll test this then merge today

ZeroChaos-:
fixing a few other shellcheck warnings while I tested this, so I'm committing my version which includes your fix.  thanks!

blshkv:
I think an another option is to change shebang to /bin/bash.

PR #277, opened 2018-04-06
Update pentoo-system-2018.0-r8.ebuild

necrose99:
sys-kernel/genkernel-next , adds a few features over reg genkernel , plymouth, at onepoint dracut ,  etc.. UEFI w/o pulling teeth    , allows either  to be used for now  https://bugs.gentoo.org/468942   features should be backported into genkernel... someday...  :-)
 a few genkernel wrappers use it , ie @sakaki- 's https://github.com/sakaki-/buildkernel for UEFI , also slightly useful for experimental porting of Pentoo to RPI3 or other arm/ amr64 boards.

### DIFF
--- a/pentoo/pentoo-system/pentoo-system-2018.0-r8.ebuild
+++ b/pentoo/pentoo-system/pentoo-system-2018.0-r8.ebuild
@@ -43,7 +43,7 @@ PDEPEND="${PDEPEND}
 	|| ( sys-process/fcron virtual/cron )
 	sys-apps/gptfdisk
 	sys-apps/pcmciautils
-	!arm? ( !livecd-stage1? ( sys-kernel/genkernel
+	!arm? ( !livecd-stage1? ( || ( sys-kernel/genkernel sys-kernel/genkernel-next )
 		|| ( sys-boot/grub:2 sys-boot/grub:0 sys-boot/grub-static sys-boot/systemd-boot )
 		)
 		sys-boot/os-prober


ZeroChaos-:
I directly develop on genkernel all the time, so odd of genkernel-next ever being the default are approximately zero.  I don't see any advantage to supporting this.  uefi works fine already and I've never found a useful feature in genkernel-next. Going to pass unless you can highlight some killer feature I'm missing.

blshkv:
I don't see why we should not give this alternative option to users even if you develop the main  tool.

necrose99:
ok np , I can add **local overlay**  to _**squash on  sakaki's  ARM64 Gentoo BASE image to shut it up... on deps**_ mainly. 
other than [sys-boot/plymouth](https://packages.gentoo.org/packages/sys-boot/plymouth) , and a few others..   , and previously Dracut , seems like most items of the fork are being ported into genkernel  anyhow.  either that or hack the ebuild buildkernel from rpi3/etc.  to use either, lob into own necromancy-overlay  

for now my sync of pentoo-overlay  on arm64 test- branch is a bit stale , https://www.pine64.org/?product=rock64-media-board-computer  after I get a board or 2 I can slowly start marking ebuilds with keywords ie ~arm64 as they build/test , with 4 gigs should be far faster... also uses EMMC so FS wont de-stabilize ...   however getting ready to move to bigger APT 3bdr apt ... for office space etc... in a few weeks.

blshkv:
Merged because it costs us nothing

blshkv:
@ZeroChaos- just found your tweet by an accident:
```
Replying to @chrisfall @pentoo_linux
Always. We accept PRs on GitHub and chat in #pentoo on freenode
```

yes we do ;-)

Issue #278, opened 2018-05-06
net-wireless/op25 fails to build due to cmake/Python-related problems


WGH-:
The basic symptom is that cmake says
```
-- Could NOT find PythonLibs (missing: PYTHON_LIBRARIES PYTHON_INCLUDE_DIRS) (Required is at least version "2")
```
and then proceeds to use plain `python`, which usually is `python3` on modern systems, which causes prompt syntax errors and build failure.

Interestingly enough, this does not happen when you clone the repository and build the code outside ebuild system.

I compared cmake traces. and found what causes the problem.

Apparently, when cmake is running under ebuild, it behaves differently with regard to python versions. There is this Gentoo-specific patch in mainline Portage tree (`/usr/portage/dev-util/cmake/files/cmake-3.9.0_rc2-FindPythonInterp.patch`):
```diff
From 325652950ecc5d72f0a6ac843889a048fe9e32cf Mon Sep 17 00:00:00 2001
From: Lars Wendler <polynomial-c@gentoo.org>
Date: Fri, 9 Jun 2017 02:08:03 +0200
Subject: [PATCH] Don't mess with python versions too much

---
 Modules/FindPythonInterp.cmake | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)

diff --git a/Modules/FindPythonInterp.cmake b/Modules/FindPythonInterp.cmake
index 64b98a879..73bd35b56 100644
--- a/Modules/FindPythonInterp.cmake
+++ b/Modules/FindPythonInterp.cmake
@@ -64,6 +64,11 @@ if(PythonInterp_FIND_VERSION)
 else()
     set(_PYTHON_FIND_OTHER_VERSIONS ${_PYTHON3_VERSIONS} ${_PYTHON2_VERSIONS} ${_PYTHON1_VERSIONS})
 endif()
+
+if (CMAKE_GENTOO_BUILD OR CMAKE_BUILD_TYPE STREQUAL Gentoo)
+    set(_Python_NAMES python)
+endif()
+
 find_program(PYTHON_EXECUTABLE NAMES ${_Python_NAMES})
 
 # Set up the versions we know about, in the order we will search. Always add
@@ -88,12 +93,13 @@ unset(_PYTHON2_VERSIONS)
 unset(_PYTHON3_VERSIONS)
 
 # Search for newest python version if python executable isn't found
-if(NOT PYTHON_EXECUTABLE)
[...]

PR #279, opened 2018-05-06
op25: fix build failure due to 'inherit python-r1' instead of python-single-r1 (#278)

WGH-:


blshkv:
looks good, thank you

PR #286, opened 2018-06-16
sys-power/acpid: bump to v2.0.29-r1 and cleanup

Wuodan:
pentoo use flag was blocking gentoos update of sys-power/acpid.

There were a bunch of unused files and small differences to upstreams ebuild files.
Upstreams v2.0.29-r1 no longer uses files in FILESDIR but rather downloads them.
So with that version only those files in FILESDIR with pentoo modifications are needed, the rest was deleted. I tried to split the steps up, thus the many commits.

Just a suggestion, as always ;)


blshkv:
I'll just merge it first with unstable keywords to see the changes

blshkv:
looks exactly like I wanted it! Thanks!

PR #294, opened 2018-07-14
Update qt to 5.9.6: Change accept_keywords

Wuodan:
qt 5.9.6 is stable in Gentoo, but causes blocked packages in Pentoo. The
reason is dev-qt/qtcharts, all other qt 5.9.6 packages, that we use, are stable in
Gentoo.

Updated accept_packages/dev-qt: Update version of qtcharts, remove all
other packages because they are stable in Gentoo.

Tested on 2018 rc 7.2


blshkv:
Looks good, thank you

PR #301, opened 2018-07-28
Accept keyword for pentoo-installer

Wuodan:
Users should update the installer before using it.
This commit masked it: e7e21d7ee710264ccb38c72c4eb9dd4f120ee7fd
and installer cannot be updated at the moment.


blshkv:
I have fixed it differently, thanks!

Wuodan:
Tnx - had that option in mind too.

ZeroChaos-:
Ugh, that was intentionally set that way. No one should be installing the
installer but me on the livecd.

blshkv:
Not true since you insist on upgrading it before running

PR #308, opened 2018-08-04
Remove x11-misc/toggle-touchpad

Wuodan:
It was added years ago 'to do me a favor'. It's just a personal script
and I have it in my own wuodan overlay. Can be removed imho.


blshkv:
Ok, thanks

PR #318, opened 2018-08-14
Cleanup obsolete packages

Wuodan:
Remove 3 dead packages
Fix SRC_URI of libmsr ... then it fails (on hardened)


blshkv:
Please do not delete dff yet. I spent way too much time fixing that ebuild, so let's keep it in the tree for now.

Wuodan:
dff is no longer removed with this PR.
Sorry, should have thought more about this, saw that you invested a lot of time.

PR #325, opened 2018-08-17
add ebuild for CubicSDR 0.2.3

tecknicaltom:
Adds an ebuild for CubicSDR 0.2.3

CubicSDR is a "Cross-Platform and Open-Source Software Defined Radio Application" (https://cubicsdr.com/).

I may not have the required dependencies and dependency use flags, but it built and worked for me. Note that 0.2.4 requires wxGTK-3.1 which is not yet in portage.


blshkv:
it would be good if you could sing your commits. I'm unable to merge it easily right now

tecknicaltom:
updated the commit to be signed. I'll send you an mp3 of me singing the commit if that's really what you wanted :)

ZeroChaos-:
Anton,

the verification only works if one of us signs the commit.  we need to pull things in manually not just merge PRs.

blshkv:
@ZeroChaos- not true because I can merge it now

blshkv:
@tecknicaltom this package requires pulseaudio (for RTAUDIO) and I do not have it in my setup. 
Can you check if it can be compiled with other audio systems such as ALSA optionally?

PR #326, opened 2018-08-17
Fix include-dir or dev-libs/libdwarf

Wuodan:
Had file collision on /usr/include/dwarf.h with dev-libs/elfutils.

blshkv maintains this package - please look at this.


blshkv:
need to verify tools that depends on it as well. ```dev-util/retdec``` is the first candidate

blshkv:
retdec does ```#include <libdwarf/dwarf.h>``` so it should be fine

PR #331, opened 2018-08-18
Adding ham2mon ebuild (needs review)

gkroon:
To be honest, I mostly created this ebuild because it seems issue #4 had been open for so long and I wanted an excuse to develop my newbie ebuild creation skills. So, in an effort to help out on the issue I hereby submit a very raw ebuild that may help with adding ham2mon to the overlay. At the moment it still fails at:
```
Traceback (most recent call last):
  File "/usr/bin/ham2mon", line 142, in <module>
    PARSER = parser.CLParser()
```
But I'm not quite happy with the `src_install(){}` anyway. I found it difficult to place the other scripts the main `ham2mon.py` script imports in the appropriate directory on the file system. Suggestions and edits are most welcome!


blshkv:
that ```python_domodule``` is tricky. It is only required when a software does not support system install (distutils). By using it, you are trying to fix this, but it also means that you have to fix a binary so that it would look for its modules in system dirs too (instead of a current folder)

net-analyzer/ham2mon/ham2mon-20170603.ebuild
```diff
@@ -0,0 +1,39 @@
...
+
+DESCRIPTION="A GNU Radio (GR) based SDR scanner with a Curses interface."
+HOMEPAGE="https://github.com/madengr/ham2mon"
+
+LICENSE="GPL-3"
+SLOT="0"
+KEYWORDS="~amd64 ~x86"
+IUSE=""
+
+RDEPEND="~dev-lang/python-2.7"
```
blshkv: python deps is adding automatically from an inherit call, so this line must be empty

gkroon:
Hi @blshkv, thanks for your review. I managed to fix the issue, hopefully in a proper way.

Basically I removed the implicit python dep due to the inherit call. I also removed the python-utils-r1 as it wasn't used, and made sure to fix `/usr/sbin/ham2mon` to look for its modules in `/usr/lib64/ham2mon/`. It also appears `ham2mon` needs root permissions to execute, otherwise it throws this warning: `RuntimeError: SDR hardware not detected or insufficient USB permissions. Try running as root.`, so I made sure the symlink ends up in `/usr/sbin`.

Please let me know if this the proper way to fix this.

blshkv:
we will improve it later if required. Cheers!

PR #332, opened 2018-08-18
improve the CubicSDR ebuild

tecknicaltom:
* adds dependency for wxGTK opengl USE
* adds USE flags and dependencies for alsa and pulseaudio


blshkv:
ok, looks good. Can you also bump it to 0.2.4? This version was out in may

tecknicaltom:
@blshkv 0.2.4 requires wxWidgets 3.1.1 [1] but portage only has wxGTK up to 3.0.4, so supporting 0.2.4 would require much more work

[1] https://cubicsdr.com/?p=190

blshkv:
https://github.com/gentoo/gentoo/pull/9359 seems very close to get merged

PR #334, opened 2018-08-19
dev-util/unicorn: Use python2 and fixes

Wuodan:
r4 failed to build with:
ERROR: Cannot use 'python', Python 2.4 or later is required.
       Note that Python 3 or later is not yet supported.
       Use --python=/path/to/python to specify a supported Python.
Fixes: make.sh in src_compile() did not forward those variables

@blshkv : this builds, r4 does not, that's all I know, merge or not at your will


blshkv:
true, python 3 support was inctroduced in qemu-2.12 only.

PR #336, opened 2018-08-20
Remove net-analyzer/inguma: Tool is dead 

Wuodan:
Homepage inguma.eu seems to be dead. Copied the only source-code that I
found 'git://git.kali.org/packages/inguma.git' to github. That's the best I could do.


blshkv:
This is the same guy who wrote bokken, https://github.com/inguma/bokken . However, he did not migrate inguma to github unlike bokken.

I see no point supporting a dead tool, please delete the ebuild instead.

Wuodan:
Even better, get rid of dead tools!
PR updated to 1 commit which deletes inguma. Also updated PR title.

Saw that sourceforge repo and wasted some time on it - but that code is even older: version 0.1.1 versus our v0.4
And sorry for all the commits, sometimes git drives me crazy!

PR #339, opened 2018-08-24
Proposing pentoo-updater to check if already synced today

gkroon:
Proposing a fix to have the pentoo-updater check if the repositories have already been synced today, to adhere to _common gentoo-netiquette_ and to avoid being placed on a _temporary ban list_ [1].

[1] https://www.gentoo.org/support/rsync-mirrors/


ZeroChaos-:
NACK.  For a few reasons.

1.) Unfortunately, emerge --sync syncs both gentoo and pentoo, and users may want and/or need to update the pentoo repo to successfully update.
2.) If you want to follow netiquette you can easily choose to sync once a day.

If portage introduces a way to sync only pentoo then I'll accept that PR.

gkroon:
Thanks for explaining! Perhaps in the mean time we should also consider adding some options to skip syncing, or to call specific functions such as `update_kernel()`?

Issue #341, opened 2018-08-27
new tool: XBruteForcer


gkroon:
https://github.com/Moham3dRiahi/XBruteForcer CMS brute forcer, or do we prefer Hydra et al more?

blshkv:
Hydra is missing templates for these cms. Feel free to create a pull request if you feel that this tool is better

blshkv:
done, thanks. Feel free to improve it further in a separate pull request

PR #342, opened 2018-08-28
adding XBruteForcer ebuild for #341

gkroon:
Used nikto ebuild as a guideline, but please let me know if I need to fix anything.


ZeroChaos-:
you can run "dos2unix" if you put it in DEPEND to fix the shebang (if you want)

same name complaint as other pr.  we don't need every binary installed twice, pick a name

gkroon:
Thanks for the dos2unix suggestion, quite useful! I also fixed the redundant symlink issue. Any more stuff you'd like me to change?

PR #344, opened 2018-08-28
could add to arm64 branch.. for now ..

necrose99:
https://www.dropbox.com/s/j1qma91hr1o7xwd/angst-0.4b-r2.tbz2?dl=0


blshkv:
Do you really need it?

blshkv:
My point is that this tool is dead. The latest release was back in 2001. I would rather delete it

necrose99:
noted i didn't think it was that bleeping old in gentoo ...  in that case i'll nuke  it..  from my chroot/repo 
 i'll just note it dose build but this point  if that 's **TUT**-est ok back in the ### tomb... 
so noted more recent tools  it was an option to put on RPI3/rock64 or etc at first consideration...

necrose99:
been compiling my ass of for arm64 so been busy getting most of pentoo even some "Historical" items built I suppose..

Issue #184, opened 2017-04-04
new tool: droopescan


blshkv:
The project is active:

https://github.com/droope/droopescan
A plugin-based scanner that aids security researchers in identifying issues with several CMSs, mainly Drupal & Silverstripe

blshkv:
Well done, thanks!

PR #346, opened 2018-08-29
Droopescan

gkroon:
I hereby submit an ebuild for #184. Please note that `dev-python/cement` needs to be unmasked as `droopescan` depends on version >=2.6.


gkroon:
I must admit I don't fully know how to. Should I add an entry in profiles/pentoo/base/package.accept_keywords/dev-python?

blshkv:
Yeap, you can add it with full version

gkroon:
Like this?

blshkv:
Cool, perfect!

PR #351, opened 2018-08-30
proposing to add new ebuilds to their accept_keywords file

gkroon:
This would help installing the new ebuilds I proposed without manually unmasking them first. But just to check, is this indeed desired?


blshkv:
Sure, why not

blshkv:
Please resolve conflict here as well

blshkv:
Hmm... Could you fix it again after i merged an another pull request please?

blshkv:
apparently, I can resolve it from a desktop. Fixed, thanks!

Issue #350, opened 2018-08-30
Proposing new category


gkroon:
I've been thinking that the _net-analyzer/linenum_ that I proposed might need to be moved as it's not a network analyser. It's a system enumeration script that runs check on the host it's invoked on. I also think none of the current categories fit such tools (except maybe for _app-misc_, but that one feels like a rubbish bin when nothing else fits, IMHO).

What do you guys think of of a separate _sys-enum_ category? I also wanted to propose ebuilds for [linuxprivchecker](https://github.com/sleventyeleven/linuxprivchecker), [Linux_Exploit_Suggester](https://github.com/InteliSecureLabs/Linux_Exploit_Suggester) and [unix-privesc-check](https://github.com/pentestmonkey/unix-privesc-check) which should end up in the same place as LinEnum. If you guys agree I'd be happy to work on this, of course.

blshkv:
How about existing app-exploits?

blshkv:
Sure, feel free to add if you use them

gkroon:
Thanks, will do.

PR #355, opened 2018-09-04
moving LinEnum to app-exploits as per #350

gkroon:
Also changed installation location. Similar to app-exploits/webshells, LinEnum is not run on the host and I therefore propose to install it in /usr/share whenever people want to upload it to a target system.


gkroon:
Sorry, I had an extra unintended ebuild in this one

PR #368, opened 2018-09-11
pyhunter

necrose99:


blshkv:
My assumption was wrong, please fixhttps://github.com/pentoo/pentoo-overlay/issues/370 asap

blshkv:
your ebuils are missing ```dev-python/xxx[${PYTHON_USEDEP}]``` python use dep flag. Please take a note

PR #373, opened 2018-09-21
fixing sonarqube's stop() function

gkroon:
The init file's stop() function didn't work and SonarQube continued to run in the background. This happens because the script that is launched, launches another wrapper that spawns SonarQube according to the system's arch (32/64 bit). I'm not sure how to solve this using a pidfile, and instead worked around it using `--name`. It's less elegant this way, I admit, but if anyone has a more elegant solution, please share.

Also, no logging was happening and I decided to remove this from the init file as well.


blshkv:
Let's merge it for now and wait for a more elegant solution of any

PR #374, opened 2018-09-24
proposing ebuild for smbmap

gkroon:
The same could technically be accomplished with nmap or metasploit, I agree. But I guess I've been accustomed to this tool when I predominantly used Kali. Could it also be merged? :-)


blshkv:
Sure :)

PR #378, opened 2018-10-09
sonarqube version bump and url change

gkroon:
Version bump and URI change. Seems like the former SRC_URIs fail because the server now demands authentication (HTTP 401 error for at least the main sonarqube zip). I then found out that Sonarqube now hosts their own binaries under their own domain. Please review and let me know if anything needs changing!


blshkv:
Looks good, thanks

PR #380, opened 2018-10-11
app-exploits/dirtycow: new package

oxr463:
Dirty COW (CVE-2016-5195) is a privilege escalation vulnerability in the Linux Kernel

https://dirtycow.ninja


blshkv:
Hi @oxr463, thank you, great work! Unfortunately,  I would like to avoid including small scripts or individual exploits. It would just take way too much time maintaining such software.

It would be good if you could port your exploit into metasploit or similar, and we will get it with a next version. Alternatively, we already have ```app-exploits/packetstormexploits-meta```. This exploit might be there too.

PR #388, opened 2018-11-13
pentoo/pentoo-installer: Add PDEPEND

Wuodan:
app-misc/jq: read/write json config-files
sys-fs/growpart: auto-grow partitions
See pentoo/pentoo-installer#45


ZeroChaos-:
ACK

PR #391, opened 2018-11-17
adding dirhunt deps keywords

gkroon:
Forgot to add keywords to properly update to new dirhunt version. :-)

PR #393, opened 2018-11-22
no latest dev-python/requests needed for dirhunt 0.6.0

gkroon:
#391: It is indeed not needed to fork and run the latest dev-python/requests. I downgraded geoip2 for this, and dirhunt still runs. Please tell me how I should fix the `dev-python/xxx[${PYTHON_USEDEP}]` issues? I'd be happy to learn and fix these as well. You can hold this pull request in the mean time until I've fixed it.


gkroon:
I see you already added these `[${PYTHON_USEDEP}]` to most of the new ebuilds. I also added a few more in the 2nd commit. Please let me know if this is more desired this way.

blshkv:
I didn't see geoip2's  requirement and it is ok if there is a strict version specified. Some authors require the latest for no reason however it might be also a real need.

Thanks!

PR #396, opened 2018-11-26
adding xsstrike

gkroon:
https://github.com/s0md3v/XSStrike

Can we add this one? I'd be happy to help maintain it, of course.


blshkv:
The project seems active. Let's add it in.

PR #401, opened 2018-12-10
fixing veil argument passing, and fixing tool output

gkroon:
Arguments can now be passed to the script (e.g. `veil --version`), and I'm also proposing to fix the tool output, as per @blshkv's https://github.com/Veil-Framework/Veil/issues/259#issuecomment-441524263.

Please review and let me know if anything needs changing before merging.


blshkv:
Can you please remove that small file from filesdir which you created with this patch and keep the previous approach?

gkroon:
Sure, like this?

blshkv:
Perfect, thanks. The upstream is not cooperative and closed couple of my bugs without any reasons. So I stopped touching this tool, but feel free to waste your time and make local changes in every new release.

gkroon:
Thanks, I noticed you did. For now my goal was to make the current version in the overlay work. We'll see how things progress from here.

PR #414, opened 2019-02-06
sqlmap version bump

gkroon:
sqlmap had a recent new release


blshkv:
Thanks for your help :)

PR #417, opened 2019-02-19
tplmap: new tool

gkroon:
I noticed your subtle hint in 83ebe90fe24c14240729d28ca1112dd722f39404, and I hope these are better written ebuilds. ;-) I'm here to learn from you guys. And to be sure, so please correct me if I'm wrong, if a python dependency is being built using (any?) python eclass, I should always suffix a given dependency with `[${PYTHON_USEDEP}]`, right?

Please let me know if anything needs editing before it can be merged.


blshkv:
I'm not an expert either ;-) But thinking logically (without checking the manual or reading the source of  classes), you should suffix it once you have any python class. The rational is for ebuild to make sure that deps support the same list of python versions. You can verify it by running repoman.

PR #427, opened 2019-03-18
app-forensics/pdf-parser: bump version

linxon:
ChangeLog:
```
  2017/01/07: V0.6.8 changed cPDFParseDictionary to handle strings () with % character
  2017/10/28: fixed bug
  2017/10/29: added # support for option -y
  2018/06/29: V0.6.9 added option --overridingfilters
  2018/10/20: added keywords to statistics
  2019/02/22: V0.7.0 added option -O --objstm to parse the stream of /ObjStm objects, inspired by a contributor wishing anonymity
  2019/03/01: V0.7.1 added ContainsName for correct keyword statistics (-a)
```


blshkv:
Hm.. the patch looks harmless, but again, could you submit your fixes to the upstream as well?  We want to avoid maintaining and patching each new version..

Thanks for the great work as always!

PR #429, opened 2019-03-26
net-misc/getsploit version bump

linxon:


blshkv:
so the upstream has accepted the patch. Perfect!

blshkv:
FYI, https://github.com/pentoo/pentoo-overlay/commit/b6e5e93b4660d27eab46d5a0e91fda13b62336e0

PR #433, opened 2019-03-28
dependency-check-bin: cleanup + minor fixes

gkroon:
needs to be run as root in order to update the CVE database in /etc/dependency-check/data


blshkv:
Please run repoman and fix tabs

blshkv:
also, there is no point in -r1 since you just increase a version

blshkv:
ok, nothing critical. Thanks!

gkroon:
> btw, it is kind of wrong to update system data, that's why it was in bin. The tool need to keep a dabase somewhere in ~./tool by right.

I'm not sure, if more users are using the system, then they will also pull in another extra ~500 MB of CVE details in all their homedirs, right?

blshkv:
Yeah, updates always tricky

PR #434, opened 2019-03-28
sonarqube: added 7.7

gkroon:
7.6 not removed since it's the last one to support 32-bit systems. I don't know why this cannot be automatically merged, though. Did I forget something?


gkroon:
Now I see, this one was already bumped :) I'm closing this one.

Issue #437, opened 2019-04-02
binwalk: add support advanced options


linxon:
Hi!

A current version of `=app-misc/binwalk-2.1.1-r1` support some [advanced options](https://github.com/ReFirmLabs/binwalk/blob/6d4483b76f916d2c993cf163836d893a3ec951df/INSTALL.md#dependencies)
I think we can expand the options if we add ebuilds for:
- ~~[jefferson](https://github.com/sviehb/jefferson)~~
- [yaffshiv](https://github.com/devttys0/yaffshiv)
- [ubi_reader](https://github.com/jrspruitt/ubi_reader)
- ~~unstuff~~

P.S.
does it really require `dev-python/pyliblzma`?
https://bugs.gentoo.org/643308

linxon:
i have old ebuild for ubi_reader in [my overlay](https://github.com/linxon/gentoo-overlay/blob/master/sys-fs/ubi_reader/ubi_reader-0.5.0.ebuild)

blshkv:
https://github.com/ReFirmLabs/binwalk/blob/master/deps.sh#L49
https://github.com/ReFirmLabs/binwalk/blob/master/deps.sh#L52
python-lzma or pyliblzma is still required unfortunately

blshkv:
I forked this tool from Gentoo because I needed squashfs to unpack a firmware. Feel free to push more modules if you need them ;-)

blshkv:
FYI, pyliblzma is not required, c02d680b4baf99573f18a8ed52c584ceedff425d

PR #438, opened 2019-04-03
app-misc/binwalk: the ebuild updates and add helpful tools

linxon:
More: https://github.com/pentoo/pentoo-overlay/issues/437


blshkv:
Great job!

PR #451, opened 2019-04-17
bulk_extractor: bump to 1.6.0_p20190416 (remove old patches)

linxon:


blshkv:
I was waiting for it! Good job with upstream!

PR #458, opened 2019-05-05
dev-util/ghidra: ebuild cleanup

jonasstein:
Add missing || die 
remove dead code
add variable ${PN}


blshkv:
Please keep the comment how to generate deps file, do not add die on "cd" command, this is really unnecessary. Btw, the next version of ghidra changed quite a bit so no point making the perfect ebuild now

blshkv:
Btw, I do not see the point of adding variables simple to replace the static names. The whole point of variables is to make it possible to copy an ebuild with a new name or make an ebuild more readable (replace long strings with MY_VAR)

blshkv:
great, thank you

blshkv:
Hi, you have asked if I know more about the coming ghidra change. It's all transparent and visible on their updated guideline:
https://github.com/NationalSecurityAgency/ghidra/blob/master/DevGuide.md

PR #463, opened 2019-05-13
wesng: 20190511 bump

gkroon:


gkroon:
@bitsadmin, which file/dir could I omit from the installation again? I remember you told me something like this a while ago when you inspected the ebuild.

bitsadmin:
That's right! CVEs.zip is only still used by legacy versions of wesng, so that one can be omitted. Definitions.zip is the new definition file format.

linxon:
thank you!

PR #477, opened 2019-06-20
gobuster: 3.0.0 bump

gkroon:


blshkv:
I hope you have tested the build and these extra deps are getting in the account?
Because usually they don't by default and it is required to create a customised src_compile function.

blshkv:
merged for testing ;-) Thanks!

gkroon:
> I hope you have tested the build and these extra deps are getting in the account?
> Because usually they don't by default and it is required to create a customised src_compile function.

I only had to add a few new deps, and remove another before it compiled successfully. I did test both the directory and dns brute force modules and they both worked fine. Of course, if anything else doesn't work, then I'd be happy to update the ebuild again.

PR #479, opened 2019-07-06
pret: new tool

gkroon:
New tool I was using the other day on an assignment. Would be nice to have it here as well, right?


blshkv:
Looks good, cheers!

PR #481, opened 2019-07-11
new tool: Chaitin X-Ray community version

OwenChia:
[Chaitin Tech](https://www.chaitin.cn/en/intro)'s [X-Ray Vulnerability Scanner](https://www.chaitin.cn/en/intro) community version.


blshkv:
This is a freeware trial scanner with no source code available. The software should be treated  as a trojan horse. Do not use it.

Also, the tar-ball releases are hosted at github. This is against github policy I believe.

PR #489, opened 2019-07-25
=dev-python/service_identity-18.1.0: add python3_7 and pypy{,3} compat

matoro:
https://service-identity.readthedocs.io/en/stable/installation.html
> Python 2.7, 3.4 and later, as well as PyPy are supported.

### DIFF
--- a/dev-python/service_identity/service_identity-18.1.0.ebuild
+++ b/dev-python/service_identity/service_identity-18.1.0.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=6
 
-PYTHON_COMPAT=( python2_7 python3_{4,5,6} )
+PYTHON_COMPAT=( python2_7 python3_{4,5,6,7} pypy{,3} )
 inherit distutils-r1
 
 DESCRIPTION="Service identity verification for pyOpenSSL."


blshkv:
I noticed that service_identity was pushed to Gentoo. Let's use that version instead.

matoro:
> 
> 
> I noticed that service_identity was pushed to Gentoo. Let's use that version instead.

The in-tree version is stuck at v17, pentoo had v18.  Isn't that reason enough to keep it?

blshkv:
I didn't any package which would specifically require v18. Correct me if im wrong.

matoro:
> 
> 
> I didn't find any package which would specifically require v18. Correct me if im wrong.

I have an ebuild for synapse, the Matrix homeserver.  It requires 18.1.0.
https://github.com/matoro/overlay/blob/master/net-im/synapse/synapse-1.1.0.ebuild#L29
https://github.com/matrix-org/synapse/blob/master/synapse/python_dependencies.py#L48

Obviously I can put dev-python/service_identity in my overlay, I was just looking to contribute upstream and avoid making copies of deps where I can.

blshkv:
The same here, we are trying to avoid unnecessary copies :-) Gentoo has a better ebuild. Please file a report at b.g.o

blshkv:
There is seems a valid reason to bump it, based on the following comment: "validating SSL certs for IP addresses". Don't forget to mention it in your report

PR #493, opened 2019-08-01
Enteletaor

gkroon:
Similar to cottontail


gkroon:
Please don't merge this one yet - it's not yet working. :(

gkroon:
Builds and runs here now :)

net-analyzer/enteletaor/enteletaor.log
```diff
@@ -0,0 +1,6 @@
+[ERROR] 2019-08-01 19:01:19 - Starting Enteletaor execution
```
linxon: this should not be here :)

PR #496, opened 2019-08-12
nosqlmap: 20190709 bump

gkroon:
I could not get the current version to work and tried to fix it, but I couldn't figure out how to make it work using setup.py/distutils-r1. So, I decided to cheat and see how BlackArch is solving this. Turns out they simply copy it to /usr/share/nosqlmap and run everything from there: https://github.com/BlackArch/blackarch/blob/master/packages/nosqlmap/PKGBUILD

I decided to do the same and now it works.


blshkv:
Yeah, this is a wrong approach. Please try to fix setup.py properly. You can start with comparing the final directory structure, install with "python setup.py --user" etc...

blshkv:
Please try that fix and let me know if you still facing any problems

gkroon:
> Please try that fix and let me know if you still facing any problems

Did you check your fix, btw? All this appears to do is install documentation:

```
[...]
--- /usr/
--- /usr/share/
--- /usr/share/doc/
--- /usr/share/doc/nosqlmap-20190812/
>>> /usr/share/doc/nosqlmap-20190812/TODO.xz
>>> /usr/share/doc/nosqlmap-20190812/README.md.xz
[...]
```

It also seems that when installing with 'python setup.py install --user' it pulls in the exact dependency versions from nosqlmap's setup.py. And when using distutils-r1 in the ebuild, it expects those exact versions as well:

```
$ NoSQLMap
Traceback (most recent call last):
  File "/usr/lib/python-exec/python2.7/NoSQLMap", line 6, in <module>
    from pkg_resources import load_entry_point
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 3126, in <module>
    @_call_aside
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 3110, in _call_aside
    f(*args, **kwargs)
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 3139, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 583, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 596, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/usr/lib64/python2.7/site-packages/pkg_resources/__init__.py", line 784, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'pymongo==2.7.2' distribution was not found and is required by NoSQLMap
```

 I feel changing the exact versions in the setup.py is the first step and tried this out. But it appears then when I remove all the exact version numbers from the setup.py in the src_prepare() phase, it dtill doesn't work properly:

```
$ NoSQLMap
Traceback (most recent call last):
  File "/usr/lib/python-exec/python2.7/NoSQLMap", line 11, in <module>
    load_entry_point('NoSQLMap==0.7', 'console_scripts', 'NoSQLMap')()
```

I do not know how to fix this properly. I

blshkv:
Run '''nosqlmap''' lowcase?

blshkv:
So yeah, ```NoSQLMap``` is broken. They run NoSQLMap = nosqlmap:main but nosqlmap.py is installed in a wrong location (as run-time script). They can't do it.

blshkv:
https://github.com/codingo/NoSQLMap/issues/90
Let the upstream deal with it

PR #506, opened 2019-10-06
profile: fixing recon-ng

gkroon:
I think this was missing when recon-ng was bumped in baf98c48828d05ec2479fd61a7b7a01a51316f90


linxon:
thank you! :)

PR #519, opened 2019-11-01
webshells: 20190713 bump

gkroon:
bump, and repoman didn't like "Pentoo Authors" in the header


app-exploits/webshells/webshells-20190713.ebuild
```diff
@@ -1,11 +1,11 @@
-# Copyright 1999-2019 Pentoo Authors
```
linxon: lol :D

PR #539, opened 2019-12-20
xsstrike: 3.1.5 bump

gkroon:
Did not update keyword, as it's already matched with `=net-analyzer/xsstrike-3.1*`


net-analyzer/xsstrike/xsstrike-3.1.5.ebuild
```diff
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-PYTHON_COMPAT=( python3_{5,6} )
+PYTHON_COMPAT=( python3_{6,7} )
```
linxon: https://github.com/pentoo/pentoo-overlay/blob/75fa1cbd8674db5edf5916c37040315a6a2041d8/net-analyzer/xsstrike/xsstrike-3.1.4-r2.ebuild#L27-L31

unfortunately, `dev-python/fuzzywuzzy` [doesn't support](https://github.com/gentoo/gentoo/blob/371b8465ed083ea0967a010103cbb89c7ff1e753/dev-python/fuzzywuzzy/fuzzywuzzy-0.12.0.ebuild#L6) `python_targets_python3_7`
```bash
* dev-python/fuzzywuzzy
     Available versions:  
       ~    0.12.0	[PYTHON_TARGETS="python2_7 python3_5 python3_6"]	["|| ( python_targets_python2_7 python_targets_python3_5 python_targets_python3_6 )"]
```

we need update `dev-python/fuzzywuzzy` before adding py37 support for xsstrike (9999 also)

gkroon: Ah you're right, my mistake. I reverted back to `python 3_{5,6}`. Anything else you want me to change?

PR #568, opened 2020-02-16
repoman + other maintenance

gkroon:
I decided to do some maintenance, mostly repoman related. Also removed `profiles/use.local.desc` after moving some local USE flags to the packages' `metadata.xml` (see https://devmanual.gentoo.org/profiles/use.desc/index.html). I also propose to remove `cuda-rarcrypt/cuda-rarcrypt` as it's no longer in active development, cannot be merged any more due to depending on outdated CUDA, and I think hashcat and john are already great substitutes. Not yet removed, though.

I hope it is ok to have spend some time on this? Please reach out to me if anything needs to be changed before it can be merged.


app-admin/vuls/vuls-0.9.0-r1.ebuild
```diff
@@ -250,10 +253,6 @@ IUSE="policykit systemd"
 SLOT=0
```
linxon: add `RESTRICT="mirror"` at this line please

app-admin/vuls/vuls-0.9.0-r1.ebuild
```diff
@@ -262,16 +261,13 @@ RDEPEND="
 	dev-go/goval-dictionary[policykit=]
 	dev-go/gost[policykit=]
 	dev-go/go-exploitdb[policykit=]
-	policykit? ( sys-auth/polkit )
+	policykit? (
+		acct-group/vuls
+		acct-user/vuls
```
linxon: maybe it also needs changing for ebuilds of these packages
```
dev-go/go-cve-dictionary
dev-go/gost
dev-go/goval-dictionary
dev-go/go-exploitdb
dev-go/go-exploitdb
www-apps/vulsrepo
```

app-admin/vuls/vuls-0.9.0-r1.ebuild
gkroon: Added

app-admin/vuls/vuls-0.9.0-r1.ebuild
gkroon: Thanks, also changed those in 52a80a6 now

PR #572, opened 2020-02-18
gitgraber: new tool

gkroon:
New tool to scan for sensitive information within public GitHub repositories.

The only thing is that the config.py is expected to be in /usr/share/${PN}, whereas this is a typical /etc/ configuration file. Also, rawGitUrls.txt is expected to be in /usr/share/${PN}, which I think should be outputted in the current working directory. I'm not sure how to fix this properly, but maybe you have suggestions? Or maybe these are issues that should be fixed by upstream (e.g. proper setup.py, etc.)?


linxon:
nice work! Thank you

PR #576, opened 2020-02-19
repoman maintenance

gkroon:
I did some more maintenance if that's ok. Please have a look at all `metadata.xml` files, because I could not describe all local USE flags. I marked all of those with `[can anyone answer this?]`. Could anyone help with these?


linxon:
Thank you! looks fine but we need to test those packages also

P.S.
I can't install `dev-libs/libcfile` package (depend for `app-forensics/libbfio`)
We need fix these issues before merging: 
 * https://github.com/pentoo/pentoo-overlay/issues/577

gkroon:
@linxon I noticed that the USE flags I staged in the metadata.xml of app-forensics/guymanager, have been removed. Was that intentional? I also noticed that the metadata.xml of app-forensics/libpff has now been merged with my "[can anyone answer this?]" comment in there. Could we also fix that one?

PR #587, opened 2020-02-29
repoman + other maintenance

gkroon:
I did some more maintenance and replaced volatility with volatility3 which switched from py2 to py3


acct-group/thehive/thehive-0.ebuild
```diff
@@ -0,0 +1,9 @@
+# Copyright 2019-2020 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+inherit acct-group
+
+DESCRIPTION="Group for TheHive"
+ACCT_GROUP_ID=-1
```
linxon: can you check it out in your machine?
Selection_155

acct-user/cortex/cortex-0.ebuild
```diff
@@ -0,0 +1,12 @@
...
+
+EAPI=7
+
+inherit acct-user
+
+DESCRIPTION="User for Cortex"
+ACCT_USER_ID=-1
+ACCT_USER_GROUPS=( ${PN} )
+
+acct-user_add_deps
```
linxon: can you check it out in your system?

Selection_155

gkroon: That is very weird. I cannot spot any mistakes in the ebuild and I'm seeing the exact error on my system. To be honest, I cannot explain this behaviour and I've reached out on the Gentoo forums for help: https://forums.gentoo.org/viewtopic-p-8427612.html#8427612

gkroon: The plot thickens. It seems the user eclass override might have something to do with it, GDH-gentoo over at the Gentoo forums suspects this is why `-M` is not taken as a valid option, making the options 'shift' one place.

It seems the override in the Pentoo tree is also quite old, judging from the eclass header.

https://github.com/pentoo/pentoo-overlay/blob/b688d9d229fa402b35bf929a4a015333fdfc3695/eclass/user.eclass#L1

linxon:
we don’t need too many changes and commits
try to do everything with one readable commit please.

thank you

PR #652, opened 2020-07-09
net-analyzer/dirsearch: Fix dependency on python >= 3.7

s-hamann:
net-analyzer/dirsearch depends on Python with the `threads` USE flag. This flag was dropped with Python 3.7 and later. Portage therefore fails to resolve the dependency and dirsearch can not be installed for Python 3.7+.
This PR fixes the USE flag dependency so dirsearch can be installed with with modern Python versions.


blshkv:
thanks

PR #663, opened 2020-08-01
Update empire-3.3.1.ebuild

necrose99:
`# use amd64 && rm libcrypto.so.1.0.0 libssl.so.1.0.0 || die / pwsh-bin-7.0.3.ebuild`  binary pkg can be cajoled into it... some reason hates this 
a quick # out and builds.. 
and likewise can use on arm64 ,  since pwsh-bin bit messy but doable ...

### DIFF
--- a/app-exploits/empire/empire-3.3.1.ebuild
+++ b/app-exploits/empire/empire-3.3.1.ebuild
@@ -17,7 +17,7 @@ KEYWORDS="~amd64 ~x86"
 LICENSE="BSD"
 SLOT="0"
 IUSE="powershell java"
-REQUIRED_USE="powershell? ( amd64 )"
+#REQUIRED_USE="powershell? ( amd64 )"
 
 # waiting for the upstream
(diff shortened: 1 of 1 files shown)


blshkv:
this is a wrong logic. I'm pushing a better one. And no, pwsh-bin is not  messy.

necrose99:
_no biggy just for clean out local rpi refresh it was getting stuck/glitching  on that spot.._ , (app-shells/pwsh-bin)

however aside from that.. 
empire will build for it 

to empire powershell? (		powershell? ( ( app-shells/pwsh-bin ) )
		arm64? amd64? ( app-shells/pwsh-bin ) )  would likely suffice,   Locally pruned to build with powershell after re-building pwsh pkg.. 

slowly getting workable use flags for gnome/lunmina /lqxt /cinnamon , mate , etc for pentoo-desktop etc.. 
xfce is already in base image , however if i can snap a good copy , then can pub a test image,

PR #665, opened 2020-08-05
Update pentoo-cinnamon-2018.2.ebuild

necrose99:
http://gpo.zugaina.org/Overlays/sping/gnome-extra/gnome-screensaver , OLD.  #>=gnome-extra/gnome-screensaver-3.4.0 no longer in Gentoo Stale version 
https://data.gpo.zugaina.org/gentoo/gnome-extra/cinnamon-screensaver/cinnamon-screensaver-4.4.1.ebuild

### DIFF
--- a/pentoo/pentoo-cinnamon/pentoo-cinnamon-2018.2.ebuild
+++ b/pentoo/pentoo-cinnamon/pentoo-cinnamon-2018.2.ebuild
@@ -18,8 +18,9 @@ PDEPEND="
 	>=x11-terms/gnome-terminal-3.4.0
 	>=media-gfx/gnome-screenshot-3.4.1
 	>=gnome-extra/gnome-system-monitor-3.4.0
-	>=gnome-extra/gnome-screensaver-3.4.0
+	>=gnome-extra/cinnamon-screensaver-4.4.1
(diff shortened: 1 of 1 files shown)


blshkv:
I switched to KDE and nobody supports cinnamon so thanks for that.
btw, I think this is your first pull request which I can actually merge ;-)

necrose99:
FORCE built cinnamon GNOME/KDE and a few goodies for pentoo remaining. on arm64 
app-antivirus (tyton lkrg fail ) however veil /malheur so far so good.. with amd64-cross-gcc or clang-+ llvm/cross targets built.
**crossdev need to re-do 6.x 8.x gcc bit stale**. 

<b>[KAGE] (https://github.com/Zerx0r/Kage)</b> will build for pc/arm64/arm/macos etc.. with a bit of electron sorcery ... 
can rig a tarball insted of app-image/rpm etc or both. MSF gui as another option for my DYSLEXIC hide.. 

hopefully soon a kick the tires arm64 image for abusing etc.. soonish , other than ssh from pc-pentoo for recon usage should be ok. 

KDE on rpi4-4gigs prolly a pig , but 8 gigs , or if Pine64 makes a new pinebook-pro-8 gigs

PR #667, opened 2020-08-08
Update pentoo-cinnamon-2018.2.ebuild

necrose99:
Drop package due bug #508854 gcalctool-6.6.2.ebuild (dead)	 1.8	 5 years	 pacho	 Drop package due bug  )Gentoo #508854
found when force building S@#$%^ for arm64..


blshkv:
https://wiki.gnome.org/Gcalctool Renamed to Calculator

blshkv:
done, thanks!

PR #670, opened 2020-08-22
motd: Cosmetic update

Wuodan:


blshkv:
I think it's better not to add version in motd at all. bump the package and it will install the latest anyway

ZeroChaos-:
Files are versioned, it makes sure everyone who installs a versioned ebuild has exactly the same files.

blshkv:
pushed a better version

ZeroChaos-:
Way to both ignore my review and take credit for someone else's work.  Let the damn PR land next time.

blshkv:
double versioning? That's unnecessary.  It's too late anyway, I pushed a fix. Feel free to add it back if you have nothing better to do.

ZeroChaos-:
Also, for a programmer yours counting skills suck.  We have another 8 days till 2020.3

blshkv:
ZeroChaos  I know you guys still use parts of your body to mesure things, but hello, don't mess up quarters please.
January, February, and March (Q1)
April, May, and June (Q2)
July, August, and September (Q3)
October, November, and December (Q4)

ZeroChaos-:
Off by one, index by 0

blshkv:
yeah, that's messed up ;-0 Is it quarter or not? Yet, this is not a standard Gentoo format. How about YYYMMDD?

ZeroChaos-:
our livecds are all tagged 2020.0 and it doesn't confuse anyone, but sure we can count numbers like Americans if that makes you feel more comfortable

blshkv:
It is confusing for non-programmers. Zero is a point, where you haven't started counting yet (to negative or positive side). 
It is perfectly fine if it is a version like MAJOR.MINOR.PATCH. But it is clearly not, because we add -rN later.

So what would it be? A semver or a quarter? because it's kind of both now.
Let's call it YYYY.QQ.PP (where PP is patch) or stick with the proper "YYYY.quarter-rN"

PR #685, opened 2020-10-29
Removed bundle install --local from net-analyzer/wpscan

joecool1029:
Bundle check appears to be enough.
Fixes sandbox violation with >=dev-ruby/bundler-2.0.0
See: https://bugs.gentoo.org/688310#c10


ZeroChaos-:
I don't hit this error, never have.  I suspect you are hitting it because some dep is missing, this isn't the right solution.

PR #706, opened 2021-03-10
app-forensics/data-carver-processors: remove package

mm1ke:
Hi,

Don't know if this is intended, but `app-forensics/data-carver-processors` doesn't have an actual ebuild, which is why i've created this PR to simply remove the package.


ZeroChaos-:
I can't merge unsigned commits but I have made this change. Thanks for the heads up :+1:

mm1ke:
Oh, i see, usually i sign my commits, but i guess i forget here somehow.. 
Anyway, i guess the usual signing like from the gentoo main tree is enough in case i'm opening some more PRs? ;)

PR #711, opened 2021-03-30
rfhs-rfctf: add tmux to dependencies

steev:
This addresses https://github.com/rfhs/rfctf-container/issues/2


pentoo/rfhs-rfctf/rfhs-rfctf-2021.0.ebuild
```diff
@@ -13,6 +13,7 @@ IUSE="+gui +wifi"
 
 PDEPEND="
 		app-admin/supervisor
+    app-misc/tmux
```
blshkv, line 16: mssing tab here?

steev: Mmmm, probably... or more likely it was converted to spaces because that's what my editor does... lemme fix that up.

blshkv:
Please sign it, I'm unable to merge

steev:
Should be good now? or do i need to re-open because of signing with a force push?

steev:
Oh, derp, i'm an idiot, I was doing S-o-b not signing.  Sorry, too used to kernel git.

steev:
And, github won't let me put my key on the account associated with the email address because it's already on this one.  I give up, Zero can do it.

blshkv:
I was about to reject it anyway. It does not belong to that meta package. Actually, it is already part of rfctr-client.
So emerge that and enjoy ;-)

steev:
https://github.com/rfhs/rfctf-container/blob/main/contestant_containers/Dockerfile.pentoo#L12 Says it's this package for the contestant containers.  This isn't for me, was just doing it to close https://github.com/rfhs/rfctf-container/issues/2 as a favor to zero

blshkv:
emerge what you need ;-) It is also part of pentoo/pentoo-system so it should be installed if you installed Pentoo properly.

steev:
Yes, that's what this was attempting to do.  The point is to keep the containers small... not pull in everything and the kitchen sink, I think.  Again, I don't care.  Zero suggested I send it as a pull request to here, so I did, but I'm not gonna bother fighting github just to add a dep.

blshkv:
Ok, let ZC to decided to add it or not. rfhs-rfctf his baby ;-)

steev:
Sounds good to me 👍

Issue #709, opened 2021-03-30
new tool: The Fastest Hash Cracking System


blshkv:
https://github.com/HashPals/Search-That-Hash

and: https://github.com/HashPals/Name-That-Hash

PR #719, opened 2021-04-30
haiti: v1.1.1

noraj:
haiti 1.1.0 and 1.1.1 were released

- https://github.com/noraj/haiti/releases
- https://github.com/noraj/haiti/blob/master/docs/CHANGELOG.md

It now supports ruby 3.0 and a new file was added (`commons.json`)

### DIFF
--- a/app-crypt/haiti-hash/haiti-hash-1.0.1.ebuild
+++ b/app-crypt/haiti-hash/haiti-hash-1.0.1.ebuild
@@ -3,7 +3,7 @@
 
 EAPI=7
 
-USE_RUBY="ruby26 ruby27"
+USE_RUBY="ruby26 ruby27 ruby30"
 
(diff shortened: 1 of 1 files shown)


blshkv:
Thanks! FYI, there is also https://github.com/pentoo/pentoo-overlay/issues/709 (Name-That-Hash),
it would be good to evaluate which tool is better ;-)

noraj:
> 
> 
> Thanks! FYI, there is also #709 (Name-That-Hash),
> it would be good to evaluate which tool is better ;-)

I have an answer to that https://noraj.github.io/haiti/#/why
Name-That-Hash is lacking support for modern hashes and has around 280 hash types support while there are 360 in Haiti including 99% those from Name-That-Hash. Also Haiti was created 2 years before Name-That-Hash that has been created without doing a proper reconnaissance to see if other tools than hashid where existing and also because the author needed a python lib for ciphey (copycat of cyberchef).

PR #724, opened 2021-06-02
Bump two packages for the python 3.9 switch

brianmwaters:
Gentoo recently switched to using python 3.9 as the default, and these two packages were left behind on my system. I am sure there are others that need to be bumped, but this is what affected me.

This makes changes to `*-9999.ebuild` packages, which are symlinked to by actual versioned packages. I don't know what the policy is there but those packages may need to be incremented to have an `-r1` suffix.

### DIFF
--- a/app-exploits/ropper/ropper-9999.ebuild
+++ b/app-exploits/ropper/ropper-9999.ebuild
@@ -4,7 +4,7 @@
 EAPI=7
 
 DISTUTILS_USE_SETUPTOOLS=rdepend
-PYTHON_COMPAT=( python3_{6,7,8} )
+PYTHON_COMPAT=( python3_{7..9} )
 
 inherit distutils-r1
 
--- a/sys-devel/gef/gef-9999.ebuild
+++ b/sys-devel/gef/gef-9999.ebuild
@@ -3,8 +3,7 @@
 
 EAPI=7
 
-#dev-libs/keystone does not support python3.9 yet
-PYTHON_COMPAT=( python3_{7..8} )
+PYTHON_COMPAT=( python3_{7..9} )
 
 inherit eutils python-single-r1
 


blshkv:
Please sign your commits.

app-exploits/ropper/ropper-9999.ebuild
```diff
@@ -4,7 +4,7 @@
 EAPI=7
 
 DISTUTILS_USE_SETUPTOOLS=rdepend
-PYTHON_COMPAT=( python3_{6,7,8} )
+PYTHON_COMPAT=( python3_{6..9} )
```
blshkv, line 7: Please change to 7..9 to be inline

blshkv (changes requested):

app-exploits/ropper/ropper-9999.ebuild
brianmwaters: Done, thanks

blshkv: no, 7..9 please.

blshkv: no, 7..9 please.

blshkv (approved):

blshkv:
I'm still not able to merge, "The base branch requires all commits to be signed."

brianmwaters:
I don't know what I'm doing, so I just force-pushed.

Github is saying the commits are "unverified" because my key is expired, but it doesn't expire until June 20, so I'm not sure what's wrong.

Let me know if they need to be signed by a newer key and I'll jump through the hoops.

brianmwaters:
Nevermind, it's the old (pre-force push) commits that it's saying are "unverified". I don't know why it says the key is expired because they aren't signed at all. Anyway I think this is all good now.

blshkv:
done, thanks!

PR #725, opened 2021-06-03
sasquatch: support gcc 10

brianmwaters:
Sasquatch apparently depends on `-fcommon`, which was previously the default, but `-fno-common` is now the default in GCC 10. @svenschwermer figured it out in https://github.com/devttys0/sasquatch/pull/34, but @devttys0 hasn't merged it yet, so until that happens, this change simply makes the flag explicit in our ebuild.

It's hard to tell because the diff adds a new file and isn't really a "diff" per se, but the new file is just a copy of `sasquatch-20180522.ebuild` with the line

```EXTRA_CFLAGS=-fcommon \```

added, and the git commit bumped to the most recent version, for good measure.


brianmwaters:
I forgot to increment the copyright to 2021, so I just quickly force-pushed real quick to fix that.

blshkv:
done, thanks

PR #726, opened 2021-06-03
Add all optional binwalk dependencies

brianmwaters:
Binwalk has soft runtime dependencies on a bunch of miscellaneous tools (see https://github.com/ReFirmLabs/binwalk/blob/master/INSTALL.md); if binwalk needs a tool and it exists on the system, then it uses it; otherwise it prints a message to the user saying that it would be able to unpack more things if they install the tool. A lot of these tools already exist in Gentoo or Pentoo; this PR adds the handful of remaining ones that don't. It also bumps a few of the ones that pre-existed in Pentoo to support Python 3.9.

I'm not a super heavy duty binwalk user and so don't use these often, but at the very least, I can commit to leaving these installed on my system so that I will at least catch major problems like when they don't build.

Here is the list of all packages that binwalk depends on, suitable for use as a set (on my system, I called it `@binwalk`):
```app-archa/lha
app-arch/arj
app-arch/bzip2
app-arch/cabextract
app-arch/gzip
app-arch/lzop
app-arch/p7zip
app-arch/stuffit
app-arch/tar
app-forensics/sleuthkit
app-misc/binwalk
dev-embedded/srecord
sys-fs/cramfs
sys-fs/cramfsswap
sys-fs/jefferson
sys-fs/mtd-utils
sys-fs/sasquatch
sys-fs/squashfs-tools
sys-fs/ubi_reader
sys-fs/yaffshiv
```

(One minor thing for the record, not strictly related to this PR: the binwalk install readme calls for a package called `lhasa`, but Gentoo only has `lha`. According to the `lhasa` readme (https://github.com/fragglet/lhasa/blob/master/README), `lha` is an older package, and `lhasa` is supposed to be a drop-in replacement. I haven't tested it, though, but I imagine it might work.)


blshkv:
that's a nice one.

brianmwaters:
Is there any way to add the set to pentoo, so that users will have a handy `@binwalkdeps`? I only know how to do it in `/etc/profile`, but there must be a way to do it here.

blshkv:
we could have forked and ebuild and add all these options, but I don't want to maintain it ;-)
post_install says "install optional packages as needed" so users should just do that.

PR #731, opened 2021-06-23
unicorn: fix files missing from install

brianmwaters:
unicorn-1.0.3 wasn't installing any of the headers or libraries. In fact, when the python USE flag was omitted, it wasn't installing any files at all.

The problem was a minor typo, which is fixed now.


blshkv:
Nice find. I was so happy to fix a 1-year installation bug and missed it. Thanks!

PR #734, opened 2021-07-11
pwntools: set unicorn dependency to proper version

brianmwaters:
The current version of pwntools depends on a specific version of unicorn (see https://github.com/Gallopsled/pwntools/blob/5b0bec0c09c24f769a02b676500795db36dd40e0/setup.py#L63). It seems to work fine with the more recent unicorn, but according to issues on the pwntools and unicorn github (see the above link if interested), there is a bug related to MIPS support. So this just mimics pwntool's native `setup.py` by requiring a specific version of unicorn, which we already have in the pentoo tree anyway.

As with these diffs that involve new copies of files, it's hard to see a diff, but all I did was change line 35 to depend on that specific unicorn version, rather than pulling in just any version.

### DIFF
--- a/app-exploits/pwntools/pwntools-4.5.1-r1.ebuild
+++ b/app-exploits/pwntools/pwntools-4.5.1-r1.ebuild
@@ -1,0 +1,45 @@
+# Copyright 1999-2021 Gentoo Authors
+# Distributed under the terms of the GNU General Public License v2
+
+EAPI=7
+
+PYTHON_COMPAT=( python3_{7..9} )
+DISTUTILS_USE_SETUPTOOLS=rdepend
+inherit distutils-r1
+
+DESCRIPTION="CTF framework and exploit development library"
+HOMEPAGE="https://github.com/Gallopsled/pwntools"
+SRC_URI="https://github.com/Gallopsled/pwntools/archive/${PV}.tar.gz -> ${P}.tar.gz"
+
+#Most is under an MIT license, but a few pieces are under GPL or a BSD 2-clause licence
+LICENSE="MIT"
+SLOT="0"
+#wait for capstone (ROPgadget dep) to become stable
+KEYWORDS="~amd64 ~x86"
+
+RDEPEND="${PYTHON_DEPS}
+	>=dev-python/paramiko-1.15.2[${PYTHON_USEDEP}]
+	>=dev-python/mako-1.0.0[${PYTHON_USEDEP}]
+	>=dev-python/pyelftools-0.2.4[${PYTHON_USEDEP}]
+	>=dev-libs/capstone-3.0.5[python,${PYTHON_USEDEP}]
+	>=app-exploits/ropgadget-5.3[${PYTHON_USEDEP}]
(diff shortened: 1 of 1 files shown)


blshkv:
Thanks!

P.S. For the nicer "diff", you could have moved the previous to -r1 instead of coping. It is not working fully anyway.

blshkv:
hm... I looked at the pull request:
https://github.com/unicorn-engine/unicorn/pull/1243/files

it may be easier to that patch instead of going backwards.
Can you please test it?

blshkv:
```
./unicron_test2.py 
[*] '/usr/lib/python3.9/site-packages/pwnlib/data/elf/test-mips-big'
    Arch:     mips-32-big
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX disabled
    PIE:      No PIE (0x400000)
    RWX:      Has RWX segments
atoi is 0x411050
printf is 0x411070
srand is 0x411060
{'_ITM_deregisterTMCloneTable': 4264052,
 '_ITM_registerTMCloneTable': 4264012,
 '_Jv_RegisterClasses': 4264024,
 '__gmon_start__': 4264036,
 '__isoc99_scanf': 4264020,
 '__libc_start_main': 4264028,
 '__stack_chk_fail': 4264040,
 '__stack_chk_guard': 4264044,
 'atoi': 4264016,
 'printf': 4264048,
 'srand': 4264032}
```


I'm reverting your broken patch

brianmwaters:
Hm, I was lazy on this one and just copied over the version constraints I saw upstream. I didn't actually dig into it.

Mysteriously, when I try to reproduce the "bug" on the latest pwntools that we have here in pentoo (4.5.1-r2), using the scripts given in https://github.com/Gallopsled/pwntools/issues/1538#issuecomment-774631189 and https://github.com/unicorn-engine/unicorn/pull/1243#issuecomment-774630691, I get the same output regardless of what version of  Unicorn is used. And it's the allegedly buggy output that lacks the PLT debugging statements that @heapcrash reported.

```
int10h@maxwell ~/Desktop $ ./issue2.py 
[*] '/home/int10h/Desktop/test-mips-big'
    Arch:     mips-32-big
    RELRO:    Partial RELRO
    Stack:    Canary found
    NX:       NX disabled
    PIE:      No PIE (0x400000)
    RWX:      Has RWX segments
atoi is 0x411050
printf is 0x411070
srand is 0x411060
{'_ITM_deregisterTMCloneTable': 4264052,
 '_ITM_registerTMCloneTable': 4264012,
 '_Jv_RegisterClasses': 4264024,
 '__gmon_start__': 4264036,
 '__isoc99_scanf': 4264020,
 '__libc_start_main': 4264028,
 '__stack_chk_fail': 4264040,
 '__stack_chk_guard': 4264044,
 'atoi': 4264016,
 'printf': 4264048,
 'srand': 4264032}
 ```

Honestly I don't know what's going on, but if it becomes an actual issue for me I'll bring it upstream. Sorry for the bad patch.

blshkv:
hm.. the same here. I tried both rc and release, no extra DEBUG info. So RIP :-)

PR #741, opened 2021-09-19
app-antivirus/lkrg: add 0.9.1_p20210913

ajakk:
This is necessary to build against latest stable kernel (5.14, tested against 5.14.6).

```
--- lkrg-0.9.1.ebuild   2021-07-30 03:08:21.393524574 -0500
+++ lkrg-0.9.1_p20210913.ebuild 2021-09-19 17:15:37.304111225 -0500
@@ -1,4 +1,4 @@
-# Copyright 1999-2019 Gentoo Authors
+# Copyright 1999-2021 Gentoo Authors
 # Distributed under the terms of the GNU General Public License v2

 EAPI=7
@@ -7,24 +7,27 @@

 DESCRIPTION="Linux Kernel Runtime Guard"
 HOMEPAGE="https://www.openwall.com/lkrg/"
-SRC_URI="https://www.openwall.com/lkrg/${P}.tar.gz"
+COMMIT="dd7fcec11f11efe0ae2fc6b8aa7b32880484a48b"
+SRC_URI="https://github.com/openwall/lkrg/archive/${COMMIT}.tar.gz -> ${P}.tar.gz"
+#SRC_URI="https://www.openwall.com/lkrg/${P}.tar.gz"
+S="${WORKDIR}/${PN}-${COMMIT}"
+
 LICENSE="GPL-3"
 SLOT="0"
-
 KEYWORDS="~amd64 ~x86"
-IUSE=""

 MODULE_NAMES="p_lkrg(misc:${S}:${S})"
-BUILD_TARGETS="clean all"
-CONFIG_CHECK="JUMP_LABEL"

 pkg_setup() {
+       local CONFIG_CHECK="JUMP_LABEL"
        linux-mod_pkg_setup
-       #compile against selected (not running) target
+
+       # compile against selected (not running) target
        BUILD_PARAMS="P_KVER=${KV_FULL} P_KERNEL=${KERNEL_DIR}"
+       BUILD_TARGETS="all"
 }

 pkg_postinst() {
-       einfo "\nUsage:"
-       einfo "\n    ~$ modprobe p_lkrg p_init_log_level=3\n"
+       einfo "Usage:"
+       einfo "modprobe p_lkrg p_init_log_level=3"
 }
```

PR #743, opened 2021-09-25
app-antivirus/lkrg: fix broken build from modified PR

ajakk:
Signed-off-by: John Helmert III <ajak@gentoo.org>


blshkv:
yeah, thanks again. I still used "hash_commit" variable to be inline with other ebuilds, but I'll just merge your pull next time since you managed to sign it.

ajakk:
It was signed last time with the same key, you just didn't manage to verify it. GitHub shows [it as signed](https://github.com/pentoo/pentoo-overlay/pull/741/commits/16068f822ced0b42cbd9dabce872455e12c424c6), too. But I'm curious as to why you care about that anyway when you just rewrite my commits and commit them as your own.

blshkv:
The "merge" button was greyed out. That's all I know.

PR #751, opened 2021-11-07
Add airspyhf

drsn0w:
### USE flags
`udev`: install udev rules (not enabled by default but users of udev should have the `udev` USE flag enabled globally)


ZeroChaos-:
Thanks!

PR #1093, opened 2021-12-30
Maryam-9999

necrose99:
Known irritations:  Maryam-v.2.5.0. on unpack if using current versioning  $/workdir/Maryam-v.2.5.0
despite ./tags/v.${PV}.tar.gz -> ${PN}-${PV}.tar.gz"
either mv Maryam-v.2.5.0 to maryam-.2.5.0 or better force unpack to current workdir/${PN}-${PV} / etc.  dir.. 
ie Maryam-9999.ebuild  (symlink) maryam-.2.5.0.ebuild 

Maryam-v.2.5.0 has had some **3 tiny revisions** although minor / ie pypi mirrors Maryam-v.2.5.0-post1 2 3 etc 
and each rev will break things ...  so by the time you test etc a new rev.... and post-x tarball and snafu... 

been in kali/Black-Arch/etc or etc ..  (http://pkg.kali.org/pkg/maryam)  

**9999 or current E-GIT HASH git-r3**  seems safe bet / less hell  than getting jabbed over the tar-ball games.. / manifest issues  

**Egit commit** 99ae85a42b4f4468fb592538ea3186da331d1369 for " Maryam-v.2.5.0-R3 " 
package output from VMware  9999  few weeks prior ... 

[maryam_package.zip](https://github.com/pentoo/pentoo-overlay/files/7791074/maryam_package.zip)

so far runs cli etc  etc new features added quite often 

other than adding the updated requirements.txt and the new depend for web-browser services 
CLI services should function as always , i will append 1 for the [dev-python/vaderSentiment](https://github.com/cjhutto/vaderSentiment)  tomorrow.  
 needs to be packaged added to revision 3 and added recent web-browser features added 
5 new depends added to requirements.txt 

Happy holidays

### DIFF
--- a/net-analyzer/maryam - Copy/metadata.xml
+++ b/net-analyzer/maryam - Copy/metadata.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE pkgmetadata SYSTEM "http://www.gentoo.org/dtd/metadata.dtd">
+<pkgmetadata>
+	<maintainer type="person">
+		<email>unknown@pentoo.ch</email>
+		<name>Author Unknown</name>
+	</maintainer>
+</pkgmetadata>
--- a/net-analyzer/maryam/Manifest.1
+++ b/net-analyzer/maryam/Manifest.1
@@ -1,0 +1,1460 @@
+
+
+
+
+
+
+<!DOCTYPE html>
+<html lang="en" data-color-mode="auto" data-light-theme="light" data-dark-theme="dark">
+  <head>
+    <meta charset="utf-8">
+  <link rel="dns-prefetch" href="https://github.githubassets.com">
+  <link rel="dns-prefetch" href="https://avatars.githubusercontent.com">
+  <link rel="dns-prefetch" href="https://github-cloud.s3.amazonaws.com">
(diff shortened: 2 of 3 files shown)


blshkv:
seems messy and untested as usual.
the "stable" version is in the tree anyway, ```net-analyzer/maryam```

necrose99:
I tested a few weeks prior , all the pre-xmass running for the house.. 
 been , coughing up  lung with whatever going about cold wise..   last few days.. 

Yes some documented known gotchas of intelligences.. 
Tarball on mirror etc..

ZeroChaos-:
respectfully, gems like "if ... the" means there is no way this ever ran https://github.com/pentoo/pentoo-overlay/pull/1093/files#diff-0ba7dd63aaa4d02a5a7483ce8e818efbc72f10357b01cae07c347503fae611e4R33

blshkv:
I tested and pushed a new version, see https://github.com/pentoo/pentoo-overlay/commit/eabcd6b5c4152bba3402e3417eaf3e7dfa38297a

Feel free to learn from it

blshkv:
I fail to find word_cloud, but you got it right. The fix is here: https://github.com/pentoo/pentoo-overlay/commit/d276364153d03e0a8a51074a6e5c3a74d4f6b812

PR #1094, opened 2022-01-16
net-wireless/mdk4: add tool

RockyBukake:
I find that gentoo's net-wireless/mdk is very unreliable or just doesn't work so I would like to fix that with this, a _newer_ version of MDK, also would like to replace net-wireless/airgeddon mdk dependency with this one or at-least give the users a choice via a USE flag


ZeroChaos-:
Is there a reason you didn't start with the make ebuild like I suggested?  You are basically wiping years of ebuild fixes like properly setting CC

RockyBukake:
I did go that route fist but found _mdk4/src/Makefile_ very annoying to work with, I did make patch files for both Makefiles but due to my lack of experience with writing ebuilds had trouble allowing _emake -DESTDIR="${D}" install_ to install files onto the _/usr/bin/_ directory, having to patch Makefile's seemed like a bit of a hack so I wen't with this approach.

ZeroChaos-:
Dropping all the ebuild fixes is not going to be acceptable.

PR #1098, opened 2022-02-24
sys-devel/gcc-arm-none-eabi: Prefix support

Ultimator14:


ZeroChaos-:
ACK
I can't merge this without a valid gpg signature.  Would you like to sign it or should I pull manually?

ZeroChaos-:
I pulled manually and signed with my key.  Thanks for the catch!

Issue #738, opened 2021-09-03
app-admin/opensnitch: install systemd service file


telans:
Hi, it would be great if the below could be installed with the `systemd` use flag or some sort.

https://github.com/pentoo/pentoo-overlay/blob/bb5ae0d5541d1bd182a1885f34472cb7a457e686/app-admin/opensnitch/opensnitch-1.3.6.ebuild#L74

Thanks.

---

On a side note, it looks like 1.4.0 has been marked yet not tagged in git. https://github.com/evilsocket/opensnitch/commit/041cf456dc121e7871955b767f5adbf2191bf855

blshkv:
So the thing with ```systemd``` is that we don't support it. I don't have a setup to check. So please create and test required files and raise a pull request.

I subscribed to the "release" event and will be notified when it is released.

telans:
Fair enough, cheers.

PR #1106, opened 2022-05-17
app-admin/opensnitch: add use flag 'systemd' and check kernel config

kcning:
This patch adds a use flag 'systemd' that installs a systemd service file. In addition, the patched ebuild checks if the kernel config turns on the necessary option.

Close issue #738 .

Signed-off-by: Kai-Chun Ning <kaichun.ning@gmail.com>


blshkv:
/usr/lib/systemd/system/
Are you sure this is the correct path?
Im not systemd user and unable to check, but i always had an impression that it should be /etc/blah

blshkv:
Also, please sign the commit so i could merge it

kcning:
> /usr/lib/systemd/system/ Are you sure this is the correct path? Im not systemd user and unable to check, but i always had an impression that it should be /etc/blah

As far as I know /usr/lib/systemd/system/ is the dir for system units installed by the distribution package manager, while /etc/systemd/system/ is for (customized) units created by the administrator. (Please see ```man 5 systemd.unit```, Table 1).

I changed the ebuild a bit and let ```systemd_dounit``` do the job. It installs to /usr/lib/systemd/system too.

ZeroChaos-:
The systemd helper puts it in the right place.  /lib is for system managed stuff and /etc is for user added stuff.  It's a little confusing but that is how systemd does it.

blshkv:
Done, thanks!

PR #1165, opened 2022-06-23
net-analyzer/sn0int: added sn0int-0.24.2

Plexvola:
Package-Manager: Portage-3.0.30, Repoman-3.0.3

Added sn0int-0.24.2, which is just a refresh of the dependencies of the old 0.19.1 ebuild.


blshkv:
Thanks!

PR #1208, opened 2022-08-08
dev-libs/nanopb: Provides pb_release symbole

thican:
Modification required by net-wireless/qflipper

Signed-off-by: Thibaud CANALE <thican@thican.net>

PR #1209, opened 2022-08-09
net-wireless/qflipper: Calls nanopb with required symbole

thican:
Requires MR #1208, which this MR is already stacked upon.

Signed-off-by: Thibaud CANALE <thican@thican.net>


net-wireless/qflipper/files/qflipper-1.1.1_display_version.patch
```diff
@@ -0,0 +1,28 @@
...
+-!equals(HAS_COMMIT, 0) {
+-    GIT_COMMIT = unknown
+-}
++GIT_COMMIT = 9e1b07c9
+ 
+-GIT_TIMESTAMP = $$system("git log -1 --pretty=format:%ct","lines", HAS_TIMESTAMP)
+-!equals(HAS_TIMESTAMP, 0) {
+-    GIT_TIMESTAMP = 0
+-}
++GIT_TIMESTAMP = 1658935746
```
ZeroChaos-, line 25: why can't we let it autodetect so we don't have to change this patch every time?

blshkv:
can you also help to fix the issue https://github.com/pentoo/pentoo-overlay/issues/1213
since you on it?

thican:
It looks like we just forgot the dev-qt/qtconcurrent dependency.

The error message was pretty much self-explanatory:
```
flipperzero/recovery/firmwaredownloadoperation.cpp:4:10: fatal error: QtConcurrent/QtConcurrentRun: No such file or directory
    4 | #include <QtConcurrent/QtConcurrentRun>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I already have this dependency on my system, so when I test with this append-flags thanks to flag-o-matic eclass in src_configure it still works:
> append-flags -flto -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing

Should I create another commit to explicitly fix #1213?

PR #1263, opened 2022-08-18
net-wireless/qflipper: add 1.1.3, drop 1.1.1-r1

thican:
Signed-off-by: Thibaud CANALE <thican@thican.net>


ZeroChaos-:
ACK, thanks

PR #1285, opened 2022-11-17
consolidated update of dev-libs/libcdata

vwegert:


vwegert:
accident - do not merge yet!

PR #1287, opened 2022-11-20
added ebuild for dev-libs/libcaes from libyal

vwegert:


vwegert:
Please disregard the commit noise above - I've cleaned the actual contents so that only libcaes-related files are affected by this PR

PR #1372, opened 2022-12-02
added app-forensics/dfvfs

vwegert:
Added the Digital Forensics Virtual Filesystem (dfvfs) library as a prerequisite to plaso.


vwegert:
Be aware that this library needs many of the other libraries I've submitted before this...

PR #1421, opened 2022-12-14
dev-ruby/cms_scanner: update dependencies

joecool1029:
Portage dropped the old ethon version so dependency resolution was broken. I synced this with upstream. This fixes dep resolution and might be enough to allow for building wpscan on ruby3.0/3.1 target.

Signed-off-by: Joe Kappus <joe@wt.gd>


blshkv:
Looks good, thanks 👍

PR #1489, opened 2023-02-02
net-wireless/srsran :: new version

richardredditch:
new version v22.10


ZeroChaos-:
I'm going to accept this for the history, but mostly it made me work on srsRAN for a while so I'll be overwriting it pretty quick ;-)

Thanks for the push

PR #1496, opened 2023-02-13
Correct the hackrf USE flag dependency of the sdrpp pkg

Brumi-2021:
Hi,  we are having a compile problem when emerging sdrpp , with USE flag hackrf. (already reported in discord pentoo) 
The error was the following one 
image

Comparing the other sdrpp  ebuilds from Hamari overlays  to our pentoo overlays , I could find the problem in the ebuild link, 
I could not test the c/m ,  but I belived that this PR will solve both versions,  1.0.4, and -9999.
https://gpo.zugaina.org/net-wireless/sdrpp
image

Cheers,


ZeroChaos-:
any chance you want to sign this commit so I can actually merge it?  or should I just apply the changes?

PR #1498, opened 2023-03-13
net-misc/sherlock: Add dependency to live ebuild

Zer0-One:
Adds dev-python/pandas as a dependency to the live ebuild.


ZeroChaos-:
got a quick comment for why that is required you could add?

Zer0-One:
> got a quick comment for why that is required you could add?

https://github.com/sherlock-project/sherlock/blob/master/sherlock/sherlock.py#L12

Fails to execute if pandas is not available.

ZeroChaos-:
Perfect, thanks

PR #1499, opened 2023-03-16
net-misc/voipong :: ebuild update

richardredditch:
updated ebuild to emerge cleanly with no errors from git


blshkv:
The tool looks dead, the last update was 12 years ago. But ok, since you fixed it ;-)

blshkv:
apparently, your "fix" broke the ebuild. ZC reverted it https://github.com/pentoo/pentoo-overlay/commit/cd66d3e13c629c5fe75a054a3aadb85f7142ab73

PR #1504, opened 2023-04-02
net-wireless/qflipper: add 1.3.0

thican:
Also address QA warning MultipleKeywordsLine (based on `pkgcheck scan`) for stable and live ebuilds.

EDIT: fixed, no more hardcoded library path.
> I lack some knowledge about system libraries support in "pro"-suffixed files so I hardcoded path for the static library nanopb, `/usr/lib64/libprotobuf-nanopb.a`; pretty sure it will break on none amd64 architectures.


thican:
Note: updated description about the hardcoded library path.

PR #1509, opened 2023-04-17
jitsi: Add autorecording support

unqueued:
I wrote this patch years ago because I was doing a series of interviews for a podcast, and I wanted to have the option to automatically record phonecalls. The official Jitsi project is not interested in adding this feature. But I figured I would share it here.


blshkv:
sure, why not ;-)
btw, a newer version was released 2 weeks ago, feel free to bump it and adopt the patch if required.
Cheers!

unqueued:
@blshkv Sure, but it isn't just a minor version bump, they've made major changes to the build process, so I am updating that. I also have some other patches that didn't make it into production.

PR #1512, opened 2023-04-26
net-im/jitsi: Add option to not build google-contacts support

unqueued:
I copied this patch from the debian version.


unqueued:
Just noticed a few issues, I'm closing and will re-push later.

PR #1517, opened 2023-05-16
dev-ruby/cms_scanner: fix nokogiri dep

joecool1029:
Build will fail with newer unless patched.

Reference: https://github.com/wpscanteam/CMSScanner/blob/v0.13.8/cms_scanner.gemspec


blshkv:
this version requirement is unnecessarily strict. Gentoo doesn't have this old version, 1.14.2-r1 is the stable now.
So I lifted the requirements instead, f6e410805a94cc06aa8148131c92f28331fb54a8

PR #1520, opened 2023-06-08
layman deprecation notice

rodan:
layman has been deprecated by gentoo and eselect repository took it's place


ZeroChaos-:
thanks!

PR #1522, opened 2023-06-08
add ebuild for soapy-sdrplay3

rodan:
added an ebuild for Soapy SDR module for SDRPlay API v3.x

tested with an RSPdx device and gqrx-2.15.9-r2


net-wireless/soapy-sdrplay3/soapy-sdrplay3-20230421.ebuild
```diff
@@ -0,0 +1,38 @@
...
+IUSE=""
+
+RDEPEND="net-wireless/soapysdr"
+DEPEND="${RDEPEND}"
+
+BDEPEND=""
+
+REQUIRED_USE=""
+
+src_configure() {
```
ZeroChaos-, line 32: please don't define the default, this and src_install should be safe to remove

rodan: you're right, removed.

ZeroChaos-:
lgtm, thanks!

PR #1524, opened 2023-06-20
app-admin/opensnitch{-ebpf-module}: add 1.5.9

kcning:
Upstream has updated the way of compiling the ebpf module. Despite the name, it's still not a kernel module.
Therefore there seems no need to inherit ``linux-mod-r1.eclass``, as it provides little benefit and [enforces](https://gitweb.gentoo.org/repo/gentoo.git/commit/?id=d9687a4df038382187300d6f44230661ff5bc377) CONFIG_TRIM_UNUSED_KSYMS and CONFIG_MODULE to be set.

N.b. It should be possible to automatically rebuild the ebpf module by depending on ``virtual/dist-kernel``.
I can test it and submit a new PR when I have more time.


blshkv:
I guess, this is outdated request, since  you pushed 1.6.0? I'll close it for now. Let me know if i'm wrong

Issue #1540, opened 2023-12-03
=sys-devel/gcc-arm-none-eabi-13.2_p1::pentoo - Build error due to wrongly unpacked package name


iomonad:
The latest version fail to build due to bad package substitution name:

```
Calculating dependencies                    ... done!
Dependency resolution took 1.16 s.

[ebuild  N     ] sys-devel/gcc-arm-none-eabi-13.2_p1::pentoo  USE="-python3" 0 KiB

Total: 1 package (1 new), Size of downloads: 0 KiB

Would you like to merge these packages? [Yes/No]
>>> Verifying ebuild manifests
>>> Emerging (1 of 1) sys-devel/gcc-arm-none-eabi-13.2_p1::pentoo
>>> Failed to emerge sys-devel/gcc-arm-none-eabi-13.2_p1, Log file:
>>>  '/var/log/portage/sys-devel:gcc-arm-none-eabi-13.2_p1:20231203-212222.log'
>>> Jobs: 0 of 1 complete, 1 failed                 Load avg: 0.52, 0.46, 0.85
 * Package:    sys-devel/gcc-arm-none-eabi-13.2_p1:0
 * Repository: pentoo
 * USE:        abi_x86_64 amd64 elibc_glibc kernel_linux
 * FEATURES:   network-sandbox preserve-libs sandbox userpriv usersandbox
>>> Unpacking source...
>>> Unpacking arm-gnu-toolchain-13.2.rel1-x86_64-arm-none-eabi.tar.xz to /var/tmp/portage/sys-devel/gcc-arm-none-eabi-13.2_p1/work
>>> Source unpacked in /var/tmp/portage/sys-devel/gcc-arm-none-eabi-13.2_p1/work
 * ERROR: sys-devel/gcc-arm-none-eabi-13.2_p1::pentoo failed (prepare phase):
 *   The source directory '/var/tmp/portage/sys-devel/gcc-arm-none-eabi-13.2_p1/work/arm-gnu-toolchain-13.2.rel1-x86_64-arm-none-eabi' doesn't exist
 *
 * Call stack:
 *            ebuild.sh, line  780:  Called __ebuild_main 'prepare'
 *   phase-functions.sh, line 1072:  Called __dyn_prepare
 *   phase-functions.sh, line  400:  Called die
 * The specific snippet of code:
 *              die "The source directory '${S}' doesn't exist"
 *
[...]

PR #1541, opened 2023-12-03
fix(gcc-arm-none-eabi): correct archive name issue mentioned in #1540

iomonad:
Dirty fix that correct #1540


ZeroChaos- (commented):
Absolutely not

PR #1546, opened 2024-02-02
dev-util/ghidra-11.0.1: Fix spaces in SRC_URI

nertpinx:
This way the URIs are not concatenated together.


blshkv:
Can't merge it, your commit is not signed

blshkv:
your fix is not a complete btw, need to add ```FunctionID``` too. But good catch, thank you!

nertpinx:
Yeah, I fixed it locally, checked it merges and then used a wrong ebuild for the PR. Thanks for fixing that.

PR #1547, opened 2024-02-09
app-admin/opensnitch{,-ebpf-module}: add 1.6.5_p1

kcning:


blshkv:
Hi, the upstream released 1.6.5.1, it must be some urgent fix. Could you bump it instead?
Also, we have new checks, could you please fix it too

kcning:
> Hi, the upstream released 1.6.5.1, it must be some urgent fix. Could you bump it instead?

Certainly, I'll bump it.

About the error message:
```version 1.6.5: uses deprecated eclass: golang-vcs-snapshot (migrate to go-module.eclass)```
I can try to migrate to go-module.eclass, but it will take some time.

kcning:
The ebuild has been migrated to using ```go-module.eclass```. Right now, the dependency tar ball is hosted on a repo that I own. If you would like to host the tar ball on the Pentoo website (or somewhere else) instead, please go ahead. For example, if you can create another Github repo.

blshkv:
that's an interesting point about hosting. So far, we have been using ```pentoo.org/~home```, similar to Gentoo. However, it is not contributors friendly. It would be easier to have a separate github repository. @ZeroChaos- what do you think?

ZeroChaos-:
I don't mean anything specific to this PR but I am against accepting random tarballs into a git repo as a general rule.

blshkv:
give me a few minutes, I'm regenerating it. It was just easier to accept the pull and adjust it later

kcning:
Thank you for quickly merging this PR. I understand that accepting tarballs can be a security risk, because of the amount of code that needs to be reviewed. Regenerating the tarballs is indeed better, and I wouldn't need to maintain this extra repo for deps.

PR #1867, opened 2024-04-01
net-wireless/qflipper: add jpeg dependency, issue #1446

thican:
I wasn’t able to trigger the warnings reported in #1446 because I guess I already had the `jpeg` USE flag for `dev-qt/qtgui:5` package.
I also tried to compile with parameters `-Werror=implicit-function-declaration -Werror=implicit-int -Werror=incompatible-function-pointer-types` for `CFLAGS` and `CXXFLAGS`, I saw no other warnings.

Clean up previous version and unused patches, fix hardcoded version.


blshkv:
looks good. Thanks!

PR #1942, opened 2024-06-13
fix libfsntfs dependencies

ensch:
The old releases have been deleted from GitHub, so installing the packages no longer works.


blshkv:
looks great. Can you also delete old one? It would be easier to see the changes too.

PR #1952, opened 2024-06-16
fix steghide install when /bin/sh is dash

jlucas8:
When /bin/sh is symlinked to the dash shell, app-crypt/steghide compiles just fine but fails at the installation phase with the following error:
```
[...]
  /bin/sh libtool --tag=CXX   --mode=install /usr/lib/portage/python3.11/ebuild-helpers/xattr/install -c steghide '/var/tmp/portage/app-crypt/steghide-0.5.1-r1/image/usr/bin'
/bin/sh: 0: cannot open libtool: No such file
make[2]: *** [Makefile:439: install-binPROGRAMS] Error 2
[...]
```
This PR fixes it.

### DIFF
--- a/app-crypt/steghide/steghide-0.5.1-r1.ebuild
+++ b/app-crypt/steghide/steghide-0.5.1-r1.ebuild
@@ -43,5 +43,7 @@ src_compile() {
 }
 
 src_install() {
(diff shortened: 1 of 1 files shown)


blshkv:
this is all weird darwin problems. But the patch is similar with what we have in src_compile section. So let it be.

PR #1964, opened 2024-06-21
dev-ruby/cms_scanner: 0.13.9 relax publix_suffix dep

joecool1029:
dev-ruby/public_suffix:4 masked in gentoo to be dropped next month.

Relax to allow 5.0.5 as only 5.1.0 and later blocked

See-also: https://github.com/wpscanteam/CMSScanner/blob/v0.13.9/cms_scanner.gemspec


blshkv:
cool, thanks

PR #2132, opened 2024-10-23
burpsuite-2024.8.5.ebuild

van9stalker:


blshkv:
bump both burp and burpsuite-pro next time. It is that simple.

PR #2153, opened 2024-11-24
rename dep dev-python/{PyQt5 -> pyqt5} per upstream

kcning:


ZeroChaos-:
Thanks.  Our QT scripts require this to all get merged at once.  I've fixed it in a followup PR.

PR #2180, opened 2025-01-04
net-wireless/qflipper: remove Qt5 graphic environment support, and live ebuilds

thican:
Qt5 support is sun-setting, it is advised to drop it instead of providing both Qt5 and Qt6 when the latter is available.

Also live ebuild has been removed, no real use case and barely no new modification from upstream.


blshkv (approved):

ZeroChaos-:
@thican can you please hit the update button to rebase your changes?  Was should pass after that.

thican:
> @thican can you please hit the update button to rebase your changes? Was should pass after that.

Done.

PR #2216, opened 2025-01-25
ghidra: fix src url

burykinne:
Build ghidra currently is broken. Emerge cannot download hfsexplorer-0_21-bin.zip from old source.


ZeroChaos-:
The old files like like they are still there to me but I think this is better either way.

Thanks!

PR #2251, opened 2025-03-11
added desktop entries for burpsuite and modify icon burpsuite-pro

van9stalker:
I have assembled and tested this.
2025-03-11-192230_hyprshot


blshkv:
I think that icon is too small. I'll update it with icon64pro.png and icon64community.png from the jar (resources/Media)

PR #2276, opened 2025-04-21
Remove very outdated flask-restx/flask-restx as its available from Guru

martin-kokos:
It's a dependency of wifipumpkin3


blshkv:
FYI, that package may no longer required by wifipumpkin3, see Changlog:
```
### Changed
- changed: replace flask-restx to flask-restful backend api
```

Issue #1769, opened 2024-02-16
dev-python/rcssmin-1.0.6 fails to compile: rcssmin.c:1085:27: error: implicit declaration of function PyUnicode_GET_SIZE 


asarubbo:
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-python/rcssmin-1.0.6 fails to compile.
Discovered on: amd64 (internal ref: pentoo_tinderbox)
System: GCC-14-SYSTEM (https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#GCC-14)

Info about the issue:
https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#CF0014

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/asarubbo/88b94e827025936c9280d0d8fe9a2a64

ZeroChaos-:
Closed by https://github.com/pentoo/pentoo-overlay/pull/2306

PR #2306, opened 2025-05-11
bump version and add upstream

rick-gnous:
To solve #1769. Compiled and tested. I've tried to add the test flag, but failed (first time, so I'm learning!). I'll maybe test again later.


ZeroChaos-:
Thanks!
Closes: https://github.com/pentoo/pentoo-overlay/issues/1769

Issue #1578, opened 2024-02-16
dev-php/evalhook-0.1_p20231013 fails to compile: evalhook.c:69:13: error: implicit declaration of function zend_string_equals_cstr [-Wimplicit-function-declaration]


asarubbo:
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-php/evalhook-0.1_p20231013 fails to compile.
Discovered on: amd64 (internal ref: pentoo_tinderbox)
System: GCC-14-SYSTEM (https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#GCC-14)

Info about the issue:
https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#CF0002

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/asarubbo/9f387d8c180df11b9e6f17e1fe7c0383

PR #2363, opened 2025-06-02
evalhook: update to the last commit, clean ebuild

rick-gnous:
Update to the last commit to avoid the error during the compilation (#1578). I've also cleaned the ebuild file.


blshkv:
you "cleaned" ebuild by removing critical PHP-slot installation and "--enable-evalhook=yes" which is absolutely required.
Have you even tested it after that? Have you tried to deobfuscate a php script? I'm pretty sure it won't work.

rick-gnous:
I've removed this by error, however I've tested and didn't get any problem to deobfuscate a PHP script. I should have been more cautious, sorry !

Issue #1407, opened 2022-12-04
dev-python/easygui-0.98.3 does not enable tests


asarubbo:
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-python/easygui-0.98.3 does not enable tests.
Discovered on: amd64 (internal ref: pentoo_tinderbox)

NOTE:
(CLANG-STRICTER-SYSTEM) in the summary means that the bug was found on a machine that runs clang with stricter mode

This machine uses clang with CFLAGS/CXXFLAGS=-Werror=implicit-function-declaration -Werror=implicit-int -Werror=incompatible-function-pointer-types

See also:
https://archives.gentoo.org/gentoo-dev/message/dd9f2d3082b8b6f8dfbccb0639e6e240
https://wiki.gentoo.org/wiki/Modern_C_porting

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/002905a2f7a8b096324dee9ee767ee96

PR #2394, opened 2025-06-15
easygui: add documentation and restrict test

rick-gnous:
I've changed the upstream for Github, so we can generate the documentation. As it doesn't have a tag, I'm using the commit SHA to download the right version (same as the pipy).

The tests need pynput (which is in guru overlay) and even with it, it has a problem because it can't acquire a X connection (I'm using tmux and st, it's maybe coming from that ?). I've restrict them for these reasons (closing #1407).


blshkv:
yeap, looks good. Thanks

PR #2444, opened 2025-07-18
pentoo-wireless: drop unneeded rtl8812au out of tree driver

ZeroChaos-:


blshkv:
why it is unneeded?
https://github.com/aircrack-ng/rtl8812au

the last commit was just 4 months ago

ZeroChaos-:
There was no in kernel driver till recently.  This external driver is pretty trash in a lot of ways, it's not worth forcing users to install when an in kernel version exists.  The hardware is all pretty crap anyway.  Either way, in kernel is by far the better choice moving forward.

PR #2458, opened 2025-07-23
Revert "pentoo-desktop-2025.3-r2.ebuild"

ZeroChaos-:
This reverts commit e844b2789661c021903d9a0469eac9b1eda1613d.
The intent is to have both chromium and firefox on the livecd


blshkv:
Users ask for one browser only. How about adding flag for each set of browsers?

PR #2461, opened 2025-07-25
dev-util/ghidra: Added a pkg_setup to call both eclasses

kmartin36:
The pkg_setup from java-pkg-2 was being overwritten by the one from python-single-r1 and was never getting called.
This was causing problems on my system due to a java version mismatch.
Explicitly calling both java-pkg-2_pkg_setup and python-single-r1_pkg_setup fixes the problem.


blshkv:
great find, thanks!

PR #2478, opened 2025-08-04
README: `emerge --sync` to `emaint sync`

vitaly-zdanevich:
Because of deprecation

> Primary control of all sync operations has been moved from `emerge` to `emaint`. `emerge --sync` now runs the emaint sync module with the` --auto` option

See https://wiki.gentoo.org/wiki/Project:Portage/Sync#Operation


vitaly-zdanevich:
Ok, in the next pull request.

PR #2481, opened 2025-08-05
pentoo-desktop: make google-chrome the default to save 600MB

ZeroChaos-:


blshkv:
should be about the same size or even smaller. Are you sure compile flags are set properly, like debug is disabled ,etc?

ZeroChaos-:
Might have been ggdb on my system.  It also saves hours.

PR #2497, opened 2025-08-22
net-analyzer/responder: needs aioquic and netifaces

goatshriek:
Responder needs the aioquic and netifaces packages to run, but these are not listed in the ebuild, which leads to a failure when trying to run it. The new ebuild is the one I am using locally to resolve this. While both dependencies are available in the main Gentoo repository, aioquic does not have an x86 keyword, so it is missing in this ebuild as well.


blshkv:
great, thank you

PR #2514, opened 2025-08-28
dev-util/ghidra-11.4.2: fix fperms glob expansion

0xseck:
During the install phase, the ebuild called fperms on globs under
Ghidra/Debug/Debugger-*/data/{support,debugger-launchers}. If no files matched, the glob was passed literally to chmod and the build failed.

Added nullglob  so only existing *.sh scripts are passed to fperms.
Without this fix, dev-util/ghidra-11.4.2 fails to install with:

`chmod: cannot access '/var/tmp/portage/dev-util/ghidra-11.4.2/image/usr/share/ghidra/Ghidra/Debug/Debugger-*/data/support/*.sh': No such file or directory`

Tested locally, build and install now succeed.


blshkv:
I'm unable to reproduce, it is getting installed here just fine. But the suggested changes won't hurt, so let's merge this pull.

blshkv:
could you sign your commit, please?

PR #2529, opened 2025-09-13
Update PyDispatcher, add tests and remove doc flag

### LINKED ISSUE #1398: dev-python/pydispatcher-2.0.5 fails to compile (CLANG-STRICTER-SYSTEM): SyntaxError: invalid syntax
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-python/pydispatcher-2.0.5 fails to compile (CLANG-STRICTER-SYSTEM).
Discovered on: amd64 (internal ref: pentoo_tinderbox)

NOTE:
(CLANG-STRICTER-SYSTEM) in the summary means that the bug was found on a machine that runs clang with stricter mode

This machine uses clang with CFLAGS/CXXFLAGS=-Werror=implicit-function-declaration -Werror=implicit-int -Werror=incompatible-function-pointer-types

See also:
https://archives.gentoo.org/gentoo-dev/message/dd9f2d3082b8b6f8dfbccb0639e6e240
https://wiki.gentoo.org/wiki/Modern_C_porting

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/63eadeb9edc9e8e48b80b00fd066d606

rick-gnous:
The documentation is using old API and deprecated Python package. close #1398

### DIFF
--- a/dev-python/pylzma/pylzma-0.5.0.ebuild
+++ b/dev-python/pylzma/pylzma-0.5.0.ebuild
@@ -4,10 +4,9 @@
 EAPI=8
 
 DISTUTILS_USE_PEP517=setuptools
-# https://github.com/fancycode/pylzma/issues/80
-# no python 3.13 support
-PYTHON_COMPAT=( python3_{11..12} )
-inherit distutils-r1
+DISTUTILS_EXT=1
+PYTHON_COMPAT=( python3_{11..13} )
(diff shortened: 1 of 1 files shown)


blshkv:
great, thank you

PR #2545, opened 2025-10-01
Ghidra: Fix globbing when Ghidra isn't installed

nathanielhourt:
When first installing Ghidra, globbing the files in /usr/share/ghidra
doesn't work because they don't exist. Work around this issue.

This is not a duplicate of pull request 2514 and this failure occurs
even with that patch applied.


blshkv:
Please sign your commit so it could be merged automatically

ZeroChaos-:
@nathanielhourt any chance you want to add your gpg key to your github account so I can merge this?  I can also merge it manually if you prefer, but then you won't show as author.

PR #2547, opened 2025-10-02
Add nuclei-bin

Miorgg:
Sorry for the mess, new PR with good email and gpg key for sig commit.


blshkv:
metadata.xml file is missing

PR #2586, opened 2025-11-10
Add OSMOCOM pySim and its dependencies

rsa9000:
Add ebuild for pySim tool [1] and all its dependencies unavailable from the main repository.

There are no fresh pySim release, so, ebuild installs latest available version from the repository. ATM, this is commit 94811ab ("pySim-shell: allow user friendly selection of the pin type").

All tests runs smoothly except some of asn1tools: some pyparsing tests have been removed due to extreme sensitivity to the exception messages, C/Rust code generation tests have been removed due to compilation complexity, and, finally, some command line utility tests have been disabled due to a strange PromptSession issue which I do not want to fix now.

The most debatable ebuild here is asn1tools-osmocom. Essentially it is asn1tools-0.166.0 plus one change making SET and SEQUENCE ordered [2]. It is unclear what side-effects can come with this change and why OSMOCOM folks have choosen the custom fork path. We have a few options here, like collaborate with upstream to push the change there, manually patch existing asn1tools, etc. I have created a dedicated ebuild for OSMOCOM fork to keep things simple at this specific moment. Then we can see does it work and have time to talk to upstream/OSMOCOM to streamline the ebuilds tree.

1. https://osmocom.org/projects/pysim/wiki
2. https://github.com/osmocom/asn1tools/commit/44277cdb34959196f317b6a02af20a4a1c3c4e2d


rsa9000:
Yeah. I'm checking. Could you share a clue, what this error means?

rsa9000:
I'm not familiar with the commits signing. It would be great if you can merge it this time. And I will do my homework and learn how to sign following up contributions if we need them.

blshkv:
Hey @rsa9000 thanks for accepting my suggestion and pushing your work to our overlay (instead of creating a yet another overlay).

Not related to this pull, it would be also great if you could push your patches to the upstream. I know it is a lot of extra work, but it would benefit everybody in a long run. We don't wont to adjust and reapply all these patches to each new version ;-)

blshkv:
@rsa9000 no, no need to CC me. I'm only getting involved when I need a tool ;-)

PR #2623, opened 2025-12-14
WIP: update Df packages

rick-gnous:
Still in WIP, some packages can't compile without the flag nls.


blshkv:
the current versions are no longer exist, so I merged that WIP. It won't make it worst anyway. Thanks for the massive bump

PR #2751, opened 2026-05-04
srsran

blshkv:
srsran


ZeroChaos- (changes requested):
remove the keywords from srsran and this will pass QA.  You are welcome to fix it by just removing the cmake searching for boost system module, but also I don't care.

PR #2755, opened 2026-05-05
sdrtrunk

blshkv:
sdrtrunk


ZeroChaos- (changes requested):
At a minimum add a comment about why we are ignoring flags and shipping a prebuilt binary instead of building from source or using what is already on the system.

ZeroChaos- (approved):
Thanks, that seems a lot better

PR #2784, opened 2026-05-23
add test and bump to 3.1.0 apkid

rick-gnous:


rick-gnous:
better after cleaning the  cache :)

PR #2789, opened 2026-05-27
Improve Recaf ebuild

catfromplan9:


blshkv:
I'm not sure why you deleted some ~old~ existing code from the ebuild, I merged it manually. Thanks

catfromplan9:
Wdym deleted old code? I only add new ebuild and .desktop file

catfromplan9:
Oh, that bottom bit on the ebuild? It's not true, please remove it. Just read what it says, and then look what the ebuild does. It does not "download openjfx on first run" that's BS, emerge pulls it in. Maybe the ebuild was originally doing something *extremely* stupid and that got fixed but the comment was never removed.

catfromplan9:
Also the `src_unpack()` is redundant, could you use my provided ebuild instead?

blshkv:
dude, I spent hours trying to make this ebuild working, so I know how it works and what it does. If you check your ~/.openjfx, you will be very surprised (with -system-openjfx is disabled by default).
Also, src_unpack is the proper (ala-Gentoo) way. Portage used to fail trying to unpack jar file without it. Additionally, it should not be allowed to install files from DISTFILES directly. It may still fail to install with one of sandbox flags (usersandbox etc) enabled.

So next time when you create a pull, please spend 5 minutes and write a small description what you want to change and why.

blshkv:
also, please refer to https://github.com/pst-format/libpst/pull/26  if you need a good example how to fill in a pull request.

PR #2792, opened 2026-05-28
nuclei-templates new release

Miorgg:


blshkv:
For github tar-balls, it has to be "-> ${P}.**gh**.tar.gz" according to Gentoo recommendation. FYI.

PR #2797, opened 2026-06-01
repology_check.py

blshkv:
repology_check.py

### DIFF
--- a/scripts/bumpchecker/repology_check.py
+++ b/scripts/bumpchecker/repology_check.py
@@ -3,7 +3,6 @@
 """Parse Repology API and display outdated packages in Pentoo
 """
 
-import json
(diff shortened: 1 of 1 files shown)


blshkv:
● Key changes: 
  - load_api() returns data instead of using a global
  - response.json() replaces json.loads(response.text)
  - HEADERS, OUTDATED_URL, IGNORE_PACKAGES are module-level constants (uppercase)
  - remove_prefix() replaced with str.removeprefix() (Python 3.9+)
  - Guard added for empty package_version_newest / gentoo_package_version
  - package.get("categories") or [] guards against missing/null categories
  - Output uses an f-string instead of string concatenation
  - Removed import json (no longer needed)

PR #2810, opened 2026-06-03
Fix airspyhf

### LINKED ISSUE #1847: net-wireless/airspyhf-1.6.8 installs '.a' files without static-libs USE
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: net-wireless/airspyhf-1.6.8 installs '.a' files without static-libs USE.
Discovered on: amd64 (internal ref: pentoo_tinderbox)
System: GCC-14-SYSTEM (https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#GCC-14)

Info about the issue:
https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#QA0070

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/asarubbo/f08fc7a4a9cadc57a92e1480f85c1fdf

Miorgg:
Fixes #1847


blshkv:
Problems in the PR:

  1. The install_manifest.txt editing is wrong and must be removed:
  local manifest="${S}_build/install_manifest.txt"
  sed -i '/\/usr\/lib64\/libairspyhf\.a/d' "${manifest}"
    - ${S}_build is not the cmake build dir — the cmake eclass uses ${BUILD_DIR}
    - Hardcodes /usr/lib64/ — breaks on x86/arm where it's /usr/lib
    - The manifest is only used by cmake's make uninstall, which Portage never calls — this code does nothing useful
  2. Mixed indentation — udev_dorules and manifest block use 4-space indentation instead of tabs. Portage requires tabs throughout.
  3. gcc15 patch should not apply to airspyhf-9999.ebuild — live ebuilds pull current git which may already have the fix or have diverged. The
  patch only belongs in the versioned ebuild.
  4. Patch --- header has .orig suffix (airspyhf_rx.c.orig) — cosmetic but non-standard. Should be a/tools/src/airspyhf_rx.c.

  Suggested action: Request changes — the fix is on the right track but the manifest manipulation needs to be dropped and the indentation fixed
  before merging. The gcc15 patch should be removed from the 9999 ebuild.

Issue #1590, opened 2024-02-16
dev-python/mechanize-0.4.5 fails tests: AttributeError: type object MozillaCookieJar has no attribute header 


asarubbo:
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-python/mechanize-0.4.5 fails tests.
Discovered on: amd64 (internal ref: pentoo_tinderbox)
System: GCC-14-SYSTEM (https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#GCC-14)

Info about the issue:
https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#CF0015

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/asarubbo/7869a09dac84376183e7d838808e9be5
Context of error(s): https://gist.github.com/asarubbo/c1fd4b849d55eea08f2018fe5fc3b098

Issue #1587, opened 2024-02-16
dev-python/dotty-1.5 fails to compile: error in efilter setup command: install_requires must be a string or list of strings containing valid version requirement specifiers; Expected end or semicolon (after version specifier) 


asarubbo:
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-python/dotty-1.5 fails to compile.
Discovered on: amd64 (internal ref: pentoo_tinderbox)
System: GCC-14-SYSTEM (https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#GCC-14)

Info about the issue:
https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#CF0002

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/asarubbo/334222a0c8cc7194e6558fcd1df81a42

Issue #1579, opened 2024-02-16
dev-python/Protego-0.2.1 fails tests: AttributeError: TestProtego object has no attribute assertEquals Did you mean: assertEqual? 


asarubbo:
https://blogs.gentoo.org/ago/2020/07/04/gentoo-tinderbox/

Issue: dev-python/Protego-0.2.1 fails tests.
Discovered on: amd64 (internal ref: pentoo_tinderbox)
System: GCC-14-SYSTEM (https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#GCC-14)

Info about the issue:
https://wiki.gentoo.org/wiki/Project:Tinderbox/Common_Issues_Helper#CF0015

At the time of writing, github-cli does not support attachments. I'm using gists but they can be deleted at a certain point. If you want to save them, please create attachments on your own.

build log: https://gist.github.com/asarubbo/b330439f5936917dc498f87ec93606e1
Context of error(s): https://gist.github.com/asarubbo/ca04c6aef0606a320a0090e89facf38c

PR #2818, opened 2026-06-04
Fix multiples tests

rick-gnous:
the usual bump, adding test and documentation, updating metada.

two packages have been renamed:
- protego
- isc-dhcp-leases

The packages that depend on them got their dependencies updated and compiled.

remove:
- dotty: no modification since 2016, archived in 2021, nothing depends on it
- mechanize: in favor of gentoo package
- wifipumpkin3: no more version since 3 years on the open source, while the pro version is still updated. it depends on qt5, that has been masked and soon [dropped in gentoo](https://bugs.gentoo.org/show_bug.cgi?id=qt5-removal)

tested on Qemu VM with FEATURES=test

close:
- #884
- #1181
- #1590
- #1579
- #1587
- #1223


rick-gnous:
welp, i guess it's time to drop wifipumpkin3, as it's only on pyqt5. furthermore, it looks like they dropped the support for the OpenSource version and focus on pro.

blshkv:
I would remove keywords from wifipumpkin first, it doesn't hurt to keep in the tree just in case of somebody would fork it. But ok, I pushed it my overlay for now

rick-gnous:
> I would remove keywords from wifipumpkin first, it doesn't hurt to keep in the tree just in case of somebody would fork it. But ok, I pushed it my overlay for now

My bad, I'll do that next time !

PR #2854, opened 2026-06-18
Cflag issue fixed

Miorgg:
Fix for:

- https://github.com/pentoo/pentoo-overlay/issues/1844
- https://github.com/pentoo/pentoo-overlay/issues/1599
- https://github.com/pentoo/pentoo-overlay/issues/1480
- https://github.com/pentoo/pentoo-overlay/issues/1479


ZeroChaos- (changes requested):
`|| die` is required in sed calls

ZeroChaos- (commented):
The if CMakelists is also unneeded.  If that file disappears this should fail.

ZeroChaos- (approved):

PR #2857, opened 2026-06-20
fixed capstone issue

### LINKED ISSUE #2856: dev-util/radare2-6.1.6: fails to build with system capstone due to upstream bug
I tried to rebuild radare2-6.1.6 for upgrading my system using @preserved-rebuild (after a capstone update), it failed with this error:
```txt
p/riscv_cs/plugin.c:263:21: error: 'CS_MODE_RISCV_C' undeclared (first use in this function);
                       did you mean 'CS_MODE_RISCVC'?
```
After some digging, I found that this ebuild uses --with-syscapstone for compiling radare2, which is as intended by Gentoo philosophy, but that triggers an existing bug in radare2 6.1.6 `libr/arch/p/riscv_cs/plugin.c`.

They use #if CS_NEXT_VERSION >= 6 to decide whether to use CS_MODE_RISCV_C (new name) or CS_MODE_RISCVC (old name), but this fails because both capstone 6.0.0 (stable) and capstone-next (dev) define CS_NEXT_VERSION as 7, making the version guard useless.

The fix is to check for CS_VERSION_PRE_RELEASE instead, which only exists in capstone-next (where CS_MODE_RISCV_C is available), not in capstone 6.0.0 (which still uses CS_MODE_RISCVC). Tested the ebuild a couple times and installed it, all works fine. I will send a PR upstream to radare2 with the same fix, so this patch can be removed on the next version bump. (noted in a comment)

kripticni:
Closes #2856

### DIFF
--- a/dev-util/radare2/files/radare2-6.1.6-fix-riscv-cs-mode.patch
+++ b/dev-util/radare2/files/radare2-6.1.6-fix-riscv-cs-mode.patch
@@ -1,0 +1,12 @@
+--- a/libr/arch/p/riscv_cs/plugin.c
++++ b/libr/arch/p/riscv_cs/plugin.c
+@@ -259,7 +259,7 @@
+ }
+ 
+ #define CSINC RISCV
+-#if CS_NEXT_VERSION >= 6
++#ifdef CS_VERSION_PRE_RELEASE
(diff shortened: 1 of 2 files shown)


kripticni:
The build test failing at fetch phase is due to the fetch requirements and licensing, it's not related to the patch.

kripticni:
Added another patch to fix different capstone versions, more definitive comment in #2856

kripticni:
The patch passed all tests and got merged upstream.

PR #2860, opened 2026-06-21
radare2-6.1.6: Updated fix-riscv-cs-mode patch

kripticni:
Better fix for #2856.

I found an edge case in the current `radare2-6.1.6-fix-riscv-cs-mode.patch`. The `#if defined(CS_VERSION_PRE_RELEASE)` guard works on alpha7 but fails on alpha4 through alpha6 of capstone 6.0.0. 

Those versions have the same macros as alpha7 (`CS_API_MAJOR=6`, `CS_NEXT_VERSION=7`, no `CS_VERSION_PRE_RELEASE`) but still use the old enum name `CS_MODE_RISCVC`.

The fix replaces the whole `#if`/`#else`/`#endif` block with a single `#define` before the capstone include:

```c
#define CS_MODE_RISCVC CS_MODE_RISCV_C
#include <capstone/capstone.h>
```

It fixes all versions both backwards and forwards for now, for version with the incorrect enum it does the fix, and for ones with the correct enum it's just a noop.

I tested against all six capstone versions (5.0.7, alpha4 through alpha7, next). The define passes every one while the `CS_VERSION_PRE_RELEASE` guard fails on alpha7.

### DIFF
--- a/dev-util/radare2/files/radare2-6.1.6-fix-riscv-cs-mode.patch
+++ b/dev-util/radare2/files/radare2-6.1.6-fix-riscv-cs-mode.patch
@@ -1,0 +1,12 @@
+--- a/libr/arch/p/riscv_cs/plugin.c
++++ b/libr/arch/p/riscv_cs/plugin.c
+@@ -259,7 +259,7 @@
+ }
+ 
+ #define CSINC RISCV
+-#if CS_NEXT_VERSION >= 6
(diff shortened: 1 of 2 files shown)


kripticni:
@blshkv found what was causing the build failures on some capstone versions

kripticni:
Synched branches.

Issue #10, opened 2015-09-25
dev-util/jd-gui-bin missing dependency on 32bit gtk+2


tecknicaltom:
dev-util/jd-gui-bin needs a runtime dependency on x11-libs/gtk+:2[abi_x86_32]

Running the program without a 32bit gtk+2 produces the following error immediately:
```
jd-gui: error while loading shared libraries: libgtk-x11-2.0.so.0: cannot open shared object file: No such file or directory
```

It may have other dependencies, but this is the one that it's currently stopping at for me.

blshkv:
ok, fixed in commit 2fe99a534e6b3c09997d951cc18b9e912ba08b45. The tools has been released as open source as well. I will have a look and bump version 4. Cheers.

PR #2878, opened 2026-07-04
Merging PF_RING ebuilds from pkalin overlay, try 3

thinrope:
Merging PF_RING ebuilds from pkalin overlay, try 3
* signed
* only one ebuild per package
* no patches in tree


ZeroChaos- (approved):

ZeroChaos-:
Looks good to me, if the checks pass we can merge.  Thanks for the extra effort.

thinrope:
Hmm, the check failed while the kernel is not configured:
````
#10 334.5  * Found sources for kernel version:
#10 334.5  *     7.0.12-pentoo
#10 334.5  * Could not find a usable .config in the kernel source directory.
#10 334.5  * Please ensure that /usr/src/linux points to a configured set of Linux sources.
#10 334.5  * If you are using KBUILD_OUTPUT, please set the environment var so that
#10 334.5  * it points to the necessary object directory so that it might find .config.
#10 334.5  * ERROR: sys-kernel/pf_ring-kmod-9.2.0_p20260430-r1::pentoo failed (setup phase):
#10 334.5  *   Kernel not configured; no .config found in /usr/src/linux
````
is there a way around this? May be the CI/CD pipeline needs to be adjusted to have a configured kernel?

ZeroChaos-:
It's best practice in gentoo to make all kernel checks non-fatal.  Otherwise things fail on build servers, when building isos, or in CI.

thinrope:
Well, I can somehow make this non-fatal, but it will be avoiding the problem.

IMHO, the package needs configured sources, es shown:
````
#10 334.5  * Call stack:
#10 334.5  *                                ebuild.sh, line 143:  Called pkg_setup
#10 334.5  *   pf_ring-kmod-9.2.0_p20260430-r1.ebuild, line  28:  Called linux-mod-r1_pkg_setup
#10 334.5  *                      linux-mod-r1.eclass, line 346:  Called _modules_prepare_kernel
#10 334.5  *                      linux-mod-r1.eclass, line 722:  Called require_configured_kernel
#10 334.5  *                        linux-info.eclass, line 362:  Called die
#10 334.5  * The specific snippet of code:
#10 334.5  *   		die "Kernel not configured; no .config found in ${KV_OUT_DIR}"
````
How can I request the kernel to be configured in the CI/CD? I don't call any explicit `die` at this stage.

blshkv:
What's the point of the snapshot, please? I would stay on the release. Please don't delete working stable ebuilds at least

ZeroChaos-:
> How can I request the kernel to be configured in the CI/CD? I don't call any explicit `die` at this stage.

The checks for the kernel module itself can stay fatal.  I will have to see about adding a kernel config and kernel sources to the CI.  For now it's something I can manually test and approve.

thinrope:
@blshkv May be you are right, I should have left the current stable. However the pf_ring-kmod needed an update anyway.  And till now at least in my overlay I always kept the two in sync, they come from the same source tree after all.

Are you using pf_ring-kmod currently, can you test this new version?

I'll try to add 9.2.0 (as released Nov 2025) in addition to the snapshots which are what upstream considers `9.2.0-stable`. Looking at the code, there is 9.3.0 probably coming soon, so then we can drop the snapshots.

blshkv:
my don't you use our existing ebuilds and just bump it?
These new ebuilds have extra workarounds, like setting local variables for "./configure" scripts etc?
myopts has "-j1" which was already resolved and so on.
Additionally, CI may work

blshkv:
ok, I borrowed the linuxmod-r1 migration code and bumped it. Do let me know if I missed anything. Thank you!

PR #2887, opened 2026-07-11
linux-sources

blshkv:
linux-sources


ZeroChaos- (changes requested):
This completely breaks the Pentoo-lts use flag, which is still defined but not used at all.  I am pretty unhappy the CI allowed this at all.

PR #2889, opened 2026-07-11
linux-sources

blshkv:
linux-sources


ZeroChaos- (changes requested):
This is still wrong.  The way it was originally is correct.

blshkv:
> This is still wrong. The way it was originally is correct.

I know it was correct. It's just hard to understand the logic. What's wrong with the current one?

blshkv:
Specifically, that line:

https://github.com/pentoo/pentoo-overlay/commit/3919d62f89bb6304c442e93ebcb527f6993397ed#diff-b5c099f77024ea605090632dfb7d1cc5f54745da708bea2b0f30a2e2c89fcc2aR16

```pentoo? ( || ( sys-kernel/pentoo-sources sys-kernel/pentoo-lts-sources ) )```

What does it do, exactly?..

PR #2912, opened 2026-07-21
dev-python/grpcio-tools: fix protobuf dependency version

cucumberer:
grpcio-tools 1.82.1 bundles libprotoc 35.0. It generates code for protobuf 7.35. However, the ebuild requires protobuf >=6.33.5 and <7. With protobuf 6.x the generated stubs do not import. They raise google.protobuf.runtime_version.VersionError. This also breaks opensnitch-ui.

Upstream requires protobuf >=7.35.1 and <8 (see Requires-Dist). Set the same range here.


blshkv:
Could you sign your commit, please ?

PR #2918, opened 2026-07-22
update stegoveritas

rick-gnous:
fix stegoveritas-py010parser and improve ebuild
clean a little bit stegoveritas-pgp
bump stegoveritas to 1.11


blshkv:
https://github.com/pentoo/pentoo-overlay/actions/runs/29913210890/job/88900828620?pr=2918#step:8:9831

rick-gnous:
i know, i'm searching the root cause of the issue

rick-gnous:
i don't understand why it fails. it should not failed, the files are here...

rick-gnous:
it looks like the issue is coming from the ci/cd ? the function getcwd does not return an existing folder https://github.com/pentoo/pentoo-overlay/actions/runs/29937406397/job/88982413323?pr=2918#step:8:8466

blshkv:
The failure is a bug in upstream `stegoveritas.py:179`:

```python
parser.add_argument('-out', ..., default=os.path.join(os.getcwd(), 'results'))
```

Python evaluates the `default=` expression immediately when `add_argument` is called, not lazily when `-out` is absent. During `distutils-r1` test execution the process CWD is a temp path that has already been removed, so `os.getcwd()` raises `FileNotFoundError` before the test body even runs.

**Ebuild workaround** — override `python_test()` to ensure CWD is a valid directory:

```bash
python_test() {
    cd "${T}" || die
    epytest
}
```

**Proper upstream fix** — change the default to `None` and compute the path lazily:

```python
parser.add_argument('-out', metavar='dir', type=str,
    help='Directory to place output in. Defaults to ./results',
    default=None)
# ... later, when self.args.out is None:
self.args.out = os.path.join(os.getcwd(), 'results')
```

rick-gnous:
looks like it's not the only getcwd... i'll just restrict the tests for now. still strange that's happening in the container but not on my VM.

blshkv:
I fixed it here: https://github.com/pentoo/pentoo-overlay/pull/2923

Issue #300, opened 2018-07-28
dev-python/nassl-1.1.3-r1 emerge failure


bill-e-ghote:
[build.log.gz](https://github.com/pentoo/pentoo-overlay/files/2237902/build.log.gz)
[build.log.txt](https://github.com/pentoo/pentoo-overlay/files/2237903/build.log.txt)
[emerge-info.txt](https://github.com/pentoo/pentoo-overlay/files/2237904/emerge-info.txt)
[emerge-pqv.txt](https://github.com/pentoo/pentoo-overlay/files/2237905/emerge-pqv.txt)
[environment.txt](https://github.com/pentoo/pentoo-overlay/files/2237906/environment.txt)

From the build log:
<snip>
/usr/lib/gcc/x86_64-pc-linux-gnu/6.4.0/../../../../x86_64-pc-linux-gnu/bin/ld:bin/openssl-legacy/lin
ux64/libssl.a: file format not recognized; treating as linker script
/usr/lib/gcc/x86_64-pc-linux-gnu/6.4.0/../../../../x86_64-pc-linux-gnu/bin/ld:bin/openssl-legacy/lin
ux64/libssl.a:1: syntax error
collect2: error: ld returned 1 exit status
error: command 'x86_64-pc-linux-gnu-gcc' failed with exit status 1
 * ERROR: dev-python/nassl-1.1.3-r1::pentoo failed (compile phase):
</snip>
Google "file format not recognized; treating as linker script" leads to some interesting comments. Seems to be a known issue for downlevel versions.

blshkv:
upgrade to gcc 7 and try again

ZeroChaos-:
I had the same error on gcc 7, it magically resolved after trying to
upgrade world a few times.  All I know is it wasn't rebuilding openssl that
fixed it.

blshkv:
That might be the same bug as reported by Alpine https://github.com/nabla-c0d3/nassl/issues/31.
However, we do build everything from scratch using the provided script, so there is still no solution

blshkv:
can you try to rebuild the latest -r2 version please?

bill-e-ghote:
I upgraded gcc for other reasons, but the failure persists, even with the -r2 version. I did not try a world upgrade yet, as this is a slow AF Oraclebox guest instance.

blshkv:
hm.. can you try to rebuild openssl deps only?

```
RDEPEND=">=app-misc/c_rehash-1.7-r1                                                                                        
[...]

PR #2926, opened 2026-07-25
app-crypt/sirikali: add 1.8.6, restore keywords

costis-t:
## Summary

- Add `sirikali-1.8.6.ebuild` (latest upstream, 2026-03-06); 1.8.2/1.8.4 ebuilds untouched.
- Restore `KEYWORDS` (as `~amd64`) — the package has been uninstallable since 1.8.4-r1, whose keywords were commented out over upstream [#300](https://github.com/mhogomchungu/sirikali/issues/300).
- Work around #300 with a `USE=system-wallet` flag (default off → bundled lxqt_wallet, which builds; the flag keeps the external build one default-flip away when upstream fixes it).
- Fix several pre-existing ebuild issues (below).

## Status of upstream #300

Still reproduces on 1.8.6 with `lxqt-wallet-4.0.2` installed and `-DINTERNAL_LXQT_WALLET=false`. It is a **link** failure:

```
ld: undefined reference to `LXQt::Wallet::backEndIsSupported(LXQt::Wallet::BackEnd)'
ld: sirikali: protected symbol `_ZN4LXQt6Wallet12walletExistsENS0_7BackEndERK7QStringS4_' isn't defined
ld: final link failed: bad value
```

Cause at the ELF level: `liblxqt-wallet.so.6.0.0` exports these symbols as `GLOBAL PROTECTED` (`readelf -sW`), which a separately linked consumer cannot bind to. Same symptom as [lxqt/lxqt_wallet#60](https://github.com/lxqt/lxqt_wallet/issues/60). Ruled out as causes with minimal reproducers: `-fvisibility`, `-Wl,-Bsymbolic*`, `-m[no-]direct-extern-access`. Root cause still unknown; will post these findings to #300.

Reproduction gotcha: without `lxqt-wallet` installed, upstream's CMake **silently falls back** to the bundled copy — a passing build only refutes #300 if the configure log says `lxqt wallet support will be provided by an external library`.

## Changes vs 1.8.4-r1, with rationale

| Change | Reason |
|---|---|
| `USE=system-wallet` (default off) selects `INTERNAL_LXQT_WALLET` | [#300](https://github.com/mhogomchungu/sirikali/issues/300); `pkg_pretend` ewarns when enabled so nobody hits the linker error blind |
| Fix automagic `dev-libs/libpwquality` | upstream's `pkg_check_modules( PWQUALITY )` is unconditional with no CMake switch — an installed libpwquality got linked regardless of USE, an [automagic dependency](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies); now sed-ed out for `USE=-pwquality` (ldd-verified) |
| `inherit xdg` + `xdg_pkg_postinst` | `.desktop`, hicolor icons and appdata are installed but caches were never updated — [xdg.eclass](https://devmanual.gentoo.org/eclass-reference/xdg.eclass/index.html) exists for exactly this |
| `pkg_postinst`: [optfeature](https://devmanual.gentoo.org/eclass-reference/optfeature.eclass/index.html) hints for `sys-fs/cryfs`, `app-crypt/gocryptfs`, `net-fs/sshfs`, `sys-fs/fscrypt`, `sys-auth/polkit` | sirikali is a front end and installs no backend. The suggestions match the executables the engines actually invoke (`src/engines/*.cpp`: `cryfs`, `gocryptfs`, `sshfs`, `fscrypt`); polkit is for the installed `sirikali.pkexec` root-mount helper. eCryptfs is deliberately **not** suggested: the engine invokes [`ecryptfs-simple`](https://github.com/mhogomchungu/ecryptfs-simple), which `sys-fs/ecryptfs-utils` does not provide and which is not packaged; same for `encfs`, `securefs`, `cryptomator-cli`. |
| Add `keyring? ( app-crypt/libsecret )` | 1.8.4 enabled `+keyring` (`NOSECRETSUPPORT=false`) without depending on the library it links |
| Drop `USE=kwallet`; `-DNOKDESUPPORT=true` unconditionally | only [`kwallet:6`](https://packages.gentoo.org/packages/kde-frameworks/kwallet) is left in ::gentoo and upstream does not support KWallet 6 (the old `pkg_postinst` warned exactly this). KDE users are still covered via `USE=keyring`: KWallet 6 implements the Secret Service API that libsecret talks to. |
| Drop the `HAS_NETWORK_SUPPORT` sed comment | no-op in 1.8.6 — the define write is gone from CMakeLists.txt. **Heads-up:** the `#if HAS_NETWORK_SUPPORT` guards are gone too, so the update checker (`api.github.com/repos/mhogomchungu/sirikali/releases`) is now compiled in unconditionally. Dormant unless enabled (`AutoCheckForUpdates` defaults to false). I can add a patch restoring the guards if you want it excluded — your call. |
| Drop `USE=test` + `RESTRICT="!test? ( test )"` | that pair only [gates `src_test`](https://devmanual.gentoo.org/ebuild-writing/functions/src_test/index.html); upstream ships no test suite (no `enable_testing`/`add_test` in CMakeLists.txt) and no ebuild version defines `src_test`, so both lines were skeleton boilerplate with no effect |
| Drop `CMAKE_BUILD_TYPE=RELEASE` and `USE=debug` | the eclass blanks the per-type flag variables (`set(CMAKE_C/CXX_FLAGS_<TYPE> "")`, [cmake.eclass ~L669](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/cmake.eclass#n669)), so `RELEASE` changed nothing; `USE=debug` was broken in -r1 (`DCMAKE_BUILD_TYPE=` sets a shell var, not a `-D` define), and per its [global description](https://gitweb.gentoo.org/repo/gentoo.git/tree/profiles/use.desc) the flag means "extra debug codepaths", not build type — which is already user-settable via the documented [`CMAKE_BUILD_TYPE` variable](https://devmanual.gentoo.org/eclass-reference/cmake.eclass/index.html) |
| `gunzip` man pages; drop explicit `eapply_user` | simpler than the `cd`/`unpack` dance; `cmake_src_prepare` runs `default_src_prepare` ([cmake.eclass](https://gitweb.gentoo.org/repo/gentoo.git/tree/eclass/cmake.eclass): "Apply ebuild and user patches via default_src_prepare"), and the [default `src_prepare`](https://devmanual.gentoo.org/ebuild-writing/functions/src_prepare/index.html) is what calls `eapply_user` — so the explicit call ran it twice |

## Testing (amd64: GCC 15, Qt 6.11.1, binutils 2.46)

| USE | Result |
|---|---|
| `keyring pwquality` (default) | builds + installs; ldd shows libpwquality + libsecret, as intended |
| `-keyring -pwquality` | builds + installs; ldd shows **neither** (automagic fix verified) |
| `system-wallet keyring pwquality` | warns, then fails to link — #300, as documented |

Runtime, with the installed binary against `sys-fs/cryfs-1.0.3`: created a vault (xchacha20-poly1305, 1 MiB blocks), mounted via `sirikali -e -d <cipher> -z <plain> -b stdin -k rw`, `sirikali -p` lists it, sha256-verified reads, wrote through the mount, unmounted via `sirikali -u`. xdg cache updates and optfeature output confirmed in the merge log. `pkgcheck scan` from this repo checkout: nothing reported for 1.8.6 (pre-existing `RedundantVersion` for 1.8.2 left alone).

Clean-room, in addition to the host build:

- **`gentoo/stage3:latest` container** (bare `default/linux/amd64/23.0` profile): all 75 dependencies resolved from the declared DEPEND/BDEPEND alone, then both USE configurations built from source and passed the same ldd checks — i.e. the dep list is complete, not accidentally satisfied by an installed desktop.
- **`pentoolinux/pentoo-core:latest` container** (`pentoo/hardened/linux/amd64/23.0/split-usr/binary` profile, deps from your binhost): built this PR's branch checkout as `::pentoo` directly — both USE configurations compile, ldd checks pass, `sirikali --version` runs (offscreen QPA).

## Questions for the maintainer

- `~amd64` only — the only arch I could test. Want `~arm64`/`~x86` added untested?
- Happy to drop 1.8.2 (RedundantVersion) and/or the never-installable 1.8.4-r1 here or in a follow-up.
- Want the update-checker guard patch (see table)?

---

This contribution was AI-assisted; all builds, the runtime test, and the #300 reproduction were executed and verified on a real Gentoo ~amd64 system, and the ebuild was reviewed by me but I'm not that experienced.


blshkv:
the issue https://github.com/mhogomchungu/sirikali/issues/300 (https://github.com/pentoo/pentoo-overlay/pull/2927) should be resolved now, please verify and adjust ebuild if required

PR #2936, opened 2026-07-29
pentoo-misc

blshkv:
pentoo-misc

### DIFF
--- a/pentoo/pentoo-misc/metadata.xml
+++ b/pentoo/pentoo-misc/metadata.xml
@@ -8,5 +8,6 @@
 	<use>
 		<flag name="office">Install libreoffice</flag>
 		<flag name="upstream-bins">Prefer upstream bins for thunderbird</flag>
+		<flag name="vlc">Install vlc player</flag>
 	</use>
 </pkgmetadata>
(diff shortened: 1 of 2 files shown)


ZeroChaos-:
Does this cause vlc to not get pulled with the default flags?  Do we need a + here?

blshkv:
Isn't it a global flag? Should be set in profile by default

ZeroChaos-:
you think it's better to set USE=vlc in the profile instead of IUSE=+vlc in this one ebuild?  I am fine with either way, but nothing else seems to have a vlc use flag.

blshkv:
I just expected that it would be set in the Gentoo profile, and we would inherit it
That's the proper behavior for overlay user as well. 
The reason why i introduced it, is that kde users don't need vlc, for example

ZeroChaos-:
You appear to have introduced the first vlc use flag.  If I set it in the profile you still disable it the same way as a user, so it seemed clearer to me to just add the + in the ebuild.

blshkv:
hm...was it clean up from the profile? It is still listed as a global use flag, though : https://www.gentoo.org/support/use-flags/
phonon-vlc was removed [935797](https://bugs.gentoo.org/935797) and KDE Plasma no longer depends on it.
It was a mess at some point: vlc-3 was masked, version 4_pre was not stable [954006](https://bugs.gentoo.org/954006). Even today, the gui is gone, I don't want to run _pre version with some random bugs.

Shall we look for a better alternative, like mpv or mplayer? I switched to Haruna (mpv gui) because of HDR support.

ZeroChaos-:
vlc works fine.  honestly I still prefer it to mpv for most uses.  It's fine that it's optional and I think that's fine for now.  If vlc is broken in some meaningful way that I'm not aware of we can add a use flag for something else and maybe switch defaults, but afaik vlc works fine right now.

PR #2948, opened 2026-08-06
net-wireless/btscanner: bump to -r8, link wide ncurses stack (fix startup SIGSEGV)

bhyvxe:
the src_configure sed replaced -lncurses with -lncursesw -ltinfo, mixing the wide libncursesw with the non-wide libtinfo/libform/libmenu. libtinfo and libtinfow both define cur_term; the dynamic linker binds libncursesw's reference to the non-wide copy, whose TERMINAL struct layout differs, so termattrs_sp() reads a garbage offset and btscanner segfaults inside initscr() on every startup.

fix: link the whole ncurses stack as wide (-lformw -lmenuw -lncursesw -ltinfow) to match libncursesw's libtinfow dependency.

EDIT [2026-08-05]: just verified by GPG. sorry.


bhyvxe:
for added context, compiling btscanner via Pentoo on an Asus Zephyrus G14 GA401IV. I used strace which produced: 

write(1, "Opening the OUI database\nReading..." , 84) = 84
--- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=0x1e8} ---

deferencing a pointer that goes into no man's land, and there's no stack at all. so it was a garbage pointer. worth criticism: i don't have gdb. i forgor. so i had to build an extremely flimsy shim that walks the frame pointers using `glibc`'s `execinfo` to print return addresses. backtrace() needs frame pointers, so it was compiled with `-fno-omit-frame-pointer`. the process would be mid-crash i guess, so the backtracing had purpose. 

result:

termattrs_sp+0x33        
_nc_setupscreen_sp+0x40c 
newterm_sp+0xc6
newterm+0x49
initscr+0x7b
btscanner+0x87ad

btscanner calls initscr(), which goes to termattrs_sp to die. so i knew it was a system library, not btscanner. so... assembly time. `nm -D` gives `0000000000023430 T termattrs_sp`. `objdump -d --start-address=0x23430 --stop-address=0x23530 libncursesw.so.6.5` reveals a fault dereferencing the bytes stored at `0x2d0` of the TERMINAL struct `cur_term` points at. that field holds garbage (`0x120`), so the next read faults at `0x1e8`; `cur_term` was bound by the dynamic linker to the non-wide `libtinfo` copy while `libncursesw` (wide) read it with a different struct layout. non-wide interposition.

blshkv:
Nice finding. Thanks!

PR #2953, opened 2026-08-12
updates20260811

ZeroChaos-:
- **zero-profile: more ignores**
- **pentoo-source: 7.1.3->7.1.8**
- **profile: prefer system libs for nodejs**


blshkv:
>  * Node.js was built with USE='system-icu'.
>  * The JavaScript 'Temporal' API has been disabled.
>  * If you need full ES2026+ compliance, rebuild with USE='-system-icu'.

Not sure if we need it and how it could affect users

ZeroChaos-:
Fine to disable the flag if someone complains

PR #2976, opened 2026-08-26
update python packages

rick-gnous:
Update for bitstruct, xonsh, pymobiledevice3, winsspi. Add new packages, as pymobiledevice3 has new dependencies. Remove pytimeparse, as gentoo has it in their overlay; less work for us :)

I've add the dependencies in profiles, because they are unstable (for now).

all compiled and tested on qemu vm with pkg-testing-tools.


dev-python/httpx-kerberos/httpx-kerberos-0.1.2.ebuild
```diff
@@ -0,0 +1,23 @@
...
+LICENSE="MIT"
+SLOT="0"
+KEYWORDS="~amd64"
+
+RDEPEND="
+	>=dev-python/cryptography-42.0.0[${PYTHON_USEDEP}]
+	>=dev-python/pyspnego-0.10.2[${PYTHON_USEDEP}]
+	<dev-python/httpx-1.0.0[${PYTHON_USEDEP}]
+	>=dev-python/httpx-0.25.0[${PYTHON_USEDEP}]
+"
```
blshkv, line 23: The wheel build failure on Python 3.14 is caused by a missing build dependency. The `pyproject.toml` declares:

```toml
[build-system]
requires = ["hatchling", "hatch-fancy-pypi-readme"]
```

`DISTUTILS_USE_PEP517=hatchling` adds `hatchling` automatically, but `hatch-fancy-pypi-readme` (used by the hatchling metadata hook to generate the README) is not declared. Fix:

```bash
BDEPEND="dev-python/hatch-fancy-pypi-readme[${PYTHON_USEDEP}]"
```

`dev-python/hatch-fancy-pypi-readme` is already in the Gentoo tree (`hatch-fancy-pypi-readme-25.1.0.ebuild`).

rick-gnous: during my test, this dependency was pulled by httpx. i can reproduce the problem with the parameter -g in portage.

PR #2984, opened 2026-08-30
Migrate Go ebuilds to dependency tarballs

techtruth:
Migrate Go module ebuilds away from deprecated EGO_SUM/module distfile lists and older vendor tarball handling. All go-module ebuilds in the overlay now fetch dependency tarballs from pentoo-golang-dist.

Regenerate Manifests for the new dependency tarballs and clean up related Go build behavior, including switching direct go build/install calls to ego where appropriate.

Also update kcptun to use the current Go proxy source zip and build client/server directly instead of relying on the upstream release script because its been removed.

Notable changes:
- Remove EGO_SUM usage from migrated Go packages
- Add ${P}-deps.tar.xz from github.com/pentoo/pentoo-golang-dist
- Regenerate Manifests for updated source/dependency distfiles
- Remove stale network-sandbox restriction from ffuf now that deps are local
- Normalize Go invocations through ego
- Fix opensnitch build-time protobuf/protoc dependencies in BDEPEND
- Update kcptun source to the provided Go proxy pseudo-version zip


app-admin/opensnitch/opensnitch-1.8.0-r1.ebuild
```diff
@@ -22,8 +22,9 @@ KEYWORDS="amd64"
...
 REQUIRED_USE="!kernel_Hurd? ( || ( iptables nftables ) )"
 
-DEPEND=">=dev-lang/go-1.19
-	net-libs/libnetfilter_queue
+DEPEND="net-libs/libnetfilter_queue"
+BDEPEND+="
+	dev-libs/protobuf[protoc]
 	dev-go/protobuf-go
 	=dev-go/protoc-gen-go-grpc-1.3.0
 "
```
blshkv, line 33: `BDEPEND+=` works here because the `go-module` eclass pre-initializes `BDEPEND` with the Go toolchain requirement before the ebuild code runs, so this appends correctly. But it is non-obvious to readers who do not know the eclass internals. A plain `BDEPEND="..."` with an explicit `>=dev-lang/go-X.Y:=` line would be clearer.

app-admin/opensnitch/opensnitch-1.8.0-r1.ebuild
```diff
@@ -103,8 +104,8 @@ src_compile() {
 	popd > /dev/null || die
 
 	pushd daemon || die
+	go-module_src_configure
 	GOCACHE="${T}/go-cache" \
-	GOMODCACHE="${WORKDIR}/${PN}-${PV}/vendor" \
 	ego build -v -buildmode=pie -o opensnitchd || die
 	popd > /dev/null || die
```
blshkv, line 117: `go-module_src_configure` is already called automatically by portage as `src_configure` (since there is no custom `src_configure` in this ebuild). Calling it again here is redundant — it is idempotent so it will not break anything, but the extra call can be dropped.

blshkv:
@techtruth The branch protection on master requires verified (GPG-signed) commits. Your commit is currently unsigned. Could you please rebase and sign it?

```
git rebase -i origin/master   # if needed
git commit --amend --no-edit -S
git push --force
```

Make sure your GPG key is configured in Git (`git config --global commit.gpgsign true`) and associated with your GitHub account.

techtruth:
Rebased and re-signed the three commits with the matching GPG identity. Current branch history is clean now.

techtruth:
Thanks for the review. I updated the branch in response to the comments:

- Reworked the Go build dependencies so they are explicit instead of relying on `BDEPEND+=` appending to deps injected by `go-module.eclass`.
- Added `GO_OPTIONAL=1` and listed the Go-side build deps directly in `BDEPEND`, including `>=dev-lang/go-1.24.11:=` and `app-arch/unzip`, which `go-module.eclass` would normally add automatically.
- Removed the `go-module_src_configure` call from `src_compile` and moved the Go environment setup into `src_configure`, alongside `distutils-r1_src_configure`. Since `GO_OPTIONAL=1` disables `go-module`'s automatic phase handling, this keeps the Go environment setup explicit while preserving the distutils configure phase.
- Rebased and re-signed the three commits using the GPG identity matching the commit email.

PR #3002, opened 2026-09-05
app-forensics/bulk_extractor: fix doc/Diagnostic_Notes filename typo

blshkv:
The file is `Diagnostic_Notes` (not `Diagnostics_Notes`) in both 2.1.1 and 2.2.0, causing `USE=doc` installs to fail.


blshkv:
Closing — will resubmit after PR #3000 is merged.

